Issue #34, opened 2022-03-22
Epic: Configuration Extension


praveek:
### Expected Behaviour
- Migrate Configuration Extension to use the new ExtensionAPI instead of module

#### Platform and Version
Android

#### Task List
- [x] #137 
- [x] #138 
- [x] #140 
- [x] #139

PR #136, opened 2022-07-22
Configuration Extension adoption of Core 2.0

prudrabhat:
## Related Issue
Epic : #34 
-  #137
- #140 

## Motivation and Context

- ConfigurationExtension now extends Extension and uses publicly exposed services via ServiceProvider
- All of ConfigurationExtension related components are now in Kotlin.
- Features implemented:
    - Configure with AppID
    - Configure with FileAsset
    - Configure with File path
    - Update configuration
    - Clear updated configuration
    - Get configuration

- RemoteDownloader is now a public utility class.
- ConfigurationDownloader, ConfigurationRulesDownloader compose RemoteDownloader for downloads.
- Created an interface contract for a utility file caching service and made the older version of CacheManager implement it.
- Moved CacheManager to compatibility package for use by Campaign and Configuration. This is a breaking change but should be compatible with minor modifications in classes using it.
- Implemented FileUtils in kotlin to shrink exception handling. Java version can be deleted when java implementation of Configuration is deleted.
- To be implemented :
    - GetSDKIdentities
    - Registration of preprocessor
    - Bundled Rules
    - Delete old java classes and tests

## How Has This Been Tested?
make unit-test and make functional-tests

## Types of changes

## Checklist:


code/android-core-library/src/main/java/com/adobe/marketing/mobile/ConfigurationExtension.java
```diff
@@ -793,8 +794,7 @@ ConfigurationDownloader getConfigDownloader(final String appId) {
 
         // Build a Configuration Downloader
         try {
-            return new ConfigurationDownloader(getPlatformServices().getNetworkService(),
-                    getPlatformServices().getSystemInfoService(), requestUrlString);
+            return new ConfigurationDownloader(ServiceProvider.getInstance().getNetworkService(), requestUrlString);
```
spoorthipujariadobe, line 797: Don't we need to refactor this class to extend Extension instead of InternalModule? Is that happening in another PR?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/utils/RemoteDownloader.kt
```diff
@@ -0,0 +1,116 @@
...
+         * @return the metadata of the [file] if it is valid;
+         *         emptymap if metadata cannot be computed
+         */
+        fun getMetadata(file: File?): Map<String, String>
+    }
+
+    /**
+     * Allows easier mocking during tests.
+     */
+    private val downloadJobSupplier: (Networking, CacheFileService, url: String, downloadSubDirectory: String?, MetadataProvider) -> RemoteDownloadJob =
```
yangyansong-adbe, line 66: `Mockito` is updated to the latest, we can start to mock class construction with Mockito now. 
https://github.com/adobe/aepsdk-core-android/blob/791777fcddf803d7b764d37e125732b42cac706f/code/android-core-library/src/test/java/com/adobe/marketing/mobile/services/NetworkServiceTests.java#L57

code/android-core-library/src/main/java/com/adobe/marketing/mobile/utils/RemoteDownloader.kt
```diff
@@ -0,0 +1,116 @@
...
+) {
+
+    companion object {
+        private const val TAG = "RemoteDownloader"
+    }
+
+    /**
+     * Represents a component that provides metadata about file content.
+     */
+    interface MetadataProvider {
```
yangyansong-adbe, line 43: [optional] IMO, there is no harm if moving it out as a non-nested interface.

code/android-core-library/src/main/java/com/adobe/marketing/mobile/utils/RemoteDownloader.kt
```diff
@@ -0,0 +1,116 @@
...
+        companion object MetadataKeys {
```
yangyansong-adbe, line 44: [discussion] I think this is the right way to hold constants in Kotlin. At the same time, the public `RemoteDownloader` class and this public `MetadataProvider` interface may be used in other extensions' java code, so it may be worth time to discuss if we only expose public utility classes/interfaces as Java class or Java interface.
cc @praveek

code/android-core-library/src/main/java/com/adobe/marketing/mobile/utils/RemoteDownloader.kt
```diff
@@ -0,0 +1,116 @@
...
+
+        /**
+         * Retrieves metadata for the [file] provided for the purpose of
+         * conditionally fetching content from the remote url.
+         *
+         * @param file the [File] for which metadata is needed
+         * @return the metadata of the [file] if it is valid;
+         *         emptymap if metadata cannot be computed
+         */
+        fun getMetadata(file: File?): Map<String, String>
```
yangyansong-adbe, line 60: I think it might be a good idea to return an optional Map here if there is no metadata retrieved for the given `File` instance.

yangyansong-adbe: Also, the parameter could be a non-optional `File` as we have can use [@NonNull](https://developer.android.com/reference/androidx/annotation/NonNull) to the newly added public classes/interfaces.

code/android-core-library/src/main/java/com/adobe/marketing/mobile/utils/RemoteDownloader.kt
```diff
@@ -0,0 +1,116 @@
...
+     *        into which the rules must be placed
+     * @param metadataProvider the metadata provider that will be used to fetch conditional headers when making a
+     *        download request
+     * @param completionCallback the callback to notify about downloaded result
+     */
+    fun download(
+        url: String,
+        downloadSubDirectory: String?,
+        metadataProvider: MetadataProvider,
+        completionCallback: (File?) -> Unit
```
yangyansong-adbe, line 95: Can we provide another parameter for the callback function that indicates the result of downloading, for example: the network is down, URL is not valid ...., then the client can decide if they need to retry it. 
Here is an example from the Swift side:
https://github.com/adobe/aepsdk-core-ios/blob/1f037918856101e863a8e6b1bc44e04ca40f189f/AEPCore/Sources/rules/RulesDownloader.swift#L50

code/android-core-library/src/main/java/com/adobe/marketing/mobile/utils/RemoteDownloader.kt
```diff
@@ -0,0 +1,116 @@
...
+                TAG,
+                "Invalid URL: ($url). Contents cannot be downloaded."
+            )
+            completionCallback.invoke(null)
+            return
+        }
+
+        val remoteDownloadJob = downloadJobSupplier(
+            networkService,
+            cacheFileService,
```
yangyansong-adbe, line 109: I think we may not cache the downloaded files inside this class. This is a public class exposed to all extensions, if we change any caching logic inside this class, it will affect other extensions. I think it should be better If we can let clients decide how/where to cache the downloaded files.
Also, we may need to add other parameters to the `download` function that shows the metadata of the previously downloaded files and how to handle 304 (not modified) response.

code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/compatibility/CacheManager.java
```diff
@@ -0,0 +1,417 @@
...
+import java.io.File;
+import java.util.Map;
+
+/**
+ * Legacy cache file management class. Only exists to allow compatibility with
+ * Configuration and Campaign extension for using {@link com.adobe.marketing.mobile.utils.RemoteDownloader}.
+ * NOTE: Bulk of the logic in this class has been copied from
+ * {@link com.adobe.marketing.mobile.CacheManager}.
+ */
+public class CacheManager implements CacheFileService {
```
yangyansong-adbe, line 34: @praveek should we move this compatibility class to the legacy directory: https://github.com/adobe/aepsdk-core-android/tree/feature/eventhub/code/android-core-library/src/legacy?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/utility/FileUtils.kt
```diff
@@ -0,0 +1,199 @@
...
+
+    /**
+     * Verifies if the directory is writable
+     *
+     * @param directory the directory to check for
+     * @return true if [directory] represents a directory and the directory is writable.
+     *         false otherwise
+     */
+    @JvmStatic
+    fun isWritableDirectory(directory: File?): Boolean {
```
yangyansong-adbe, line 38: This is the utility class only used inside the Core project, can we make all `File` parameters non-optional and add a `@NonNull` annotation for them?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/utility/FileUtils.kt
```diff
@@ -0,0 +1,199 @@
...
+        if (!folder.exists() && !folder.mkdir()) {
+            MobileCore.log(
+                LoggingMode.WARNING,
+                TAG,
+                "Could not create the output directory $outputDirectoryPath"
+            )
+            return false
+        }
+
+        var entryProcessedSuccessfully = true
```
yangyansong-adbe, line 148: `entryProcessedSuccessfully` or `extractProcessedSuccessfully`?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/utility/FileUtils.kt
```diff
@@ -0,0 +1,199 @@
...
+                                LoggingMode.WARNING,
+                                TAG,
+                                "Could not extract the file ${newZipEntryFile.absolutePath}"
+                            )
+                            return false
+                        }
+                    }
+                    ze = zipInputStream.nextEntry
+                }
+                zipInputStream.closeEntry()
```
yangyansong-adbe, line 190: Should we call `closeEntry()` before moving to the next entry?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/utility/StringEncoder.java
```diff
@@ -96,4 +121,107 @@ private static int getFnv1aHash(final String input) {
...
+            final int radix = 16;
+            final int fourDigit = 4;
+            data[i / 2] = (byte) ((Character.digit(hexString.charAt(i), radix) << fourDigit) +
+                    Character.digit(hexString.charAt(i + 1), radix));
+        }
+
+        String decodedString = null;
+
+        try {
+            decodedString = new String(data, CHARSET_UTF_8);
```
yangyansong-adbe, line 183: Can we switch to calling another String constructor (`public String (byte[] bytes, Charset charset)`) as below?
https://developer.android.com/reference/java/lang/String?hl=en#String(byte[],%20java.nio.charset.Charset)
Also, we can remove the public constant `CHARSET_UTF_8`.
```java
decodedString = new String(data, StandardCharsets.UTF_8);
```
> **StandardCharsets** was added since Android API level 19

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/AppIdManager.kt
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * Retrieves appId from persistence when available. Falls back to
+     * retrieving it from manifest if nothing is persisted.
+     *
+     * @return the appId stored in manifest or persistence (in that order) if one exists;
+     *         null otherwise
+     */
+    internal fun loadAppId(): String? {
+        return getAppIDFromPersistence() ?: getAppIDFromManifest().also { manifestAppId ->
```
yangyansong-adbe, line 66: I'm thinking of an extreme case when the Android app is updated with a new app id in the manifest file. Should we first load the app id from the manifest?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/ConfigurationStateManager.kt
```diff
@@ -0,0 +1,417 @@
...
+     * @return the initial configuration at launch
+     */
+    internal fun loadInitialConfig(): Map<String, Any?> {
+        currentConfiguration.clear()
+
+        val appId: String? = appIdManager.loadAppId()
+
+        val config: Map<String, Any?>? = if (appId.isNullOrEmpty()) {
+            // Load bundled config
+            getBundledConfig(CONFIG_BUNDLED_FILE_NAME)
```
yangyansong-adbe, line 142: [optional] `loadBundledConfig`

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/ConfigurationStateManager.kt
```diff
@@ -0,0 +1,417 @@
...
+        } else {
+            getCachedConfig(appId) ?: getBundledConfig(CONFIG_BUNDLED_FILE_NAME)
```
yangyansong-adbe, line 144: [optional]`loadCachedConfig`

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/ConfigurationExtension.kt
```diff
@@ -0,0 +1,454 @@
...
+        super.onRegistered()
+
+        // TODO: Register pre-processor
+        api.registerEventListener(
+            EVENT_TYPE,
+            EVENT_SOURCE_REQUEST_CONTENT,
+            extensionEventListener
+        )
+
+        val appId = this.appIdManager.loadAppId()
```
yangyansong-adbe, line 157: Can we add a verbose log here if no app-id is presented in the local cache or manifest file?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/ConfigurationExtension.kt
```diff
@@ -0,0 +1,454 @@
...
+        if (!appId.isNullOrBlank()) {
+            val eventData: MutableMap<String, Any?> =
+                mutableMapOf(
+                    CONFIGURATION_REQUEST_CONTENT_JSON_APP_ID to appId,
+                    CONFIGURATION_REQUEST_CONTENT_IS_INTERNAL_EVENT to true
+                )
+            dispatchConfigurationRequest(eventData)
+        }
+
+        val initialConfig: Map<String, Any?> = this.configurationStateManager.loadInitialConfig()
```
yangyansong-adbe, line 168: Can we add a verbose log here on whether the bundled/cached config is loaded or not?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/ConfigurationExtension.kt
```diff
@@ -0,0 +1,454 @@
...
+    override fun onRegistered() {
+        super.onRegistered()
+
+        // TODO: Register pre-processor
+        api.registerEventListener(
+            EVENT_TYPE,
+            EVENT_SOURCE_REQUEST_CONTENT,
+            extensionEventListener
+        )
+
```
yangyansong-adbe, line 156: https://github.com/adobe/aepsdk-core-android/blob/791777fcddf803d7b764d37e125732b42cac706f/code/android-core-library/src/main/java/com/adobe/marketing/mobile/ConfigurationExtension.java#L246
We did loading local (bundled/cached) config when the boot event is dispatched, why do we change this logic now? Btw, I think extensions should only do `registerEventListener` operations when `onRegistered` is called.

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/ConfigurationExtension.kt
```diff
@@ -0,0 +1,454 @@
...
+    }
+
+    private val serviceProvider: ServiceProvider
+    private val appIdManager: AppIdManager
+    private val cacheFileService: CacheFileService
+    private val launchRulesEvaluator: LaunchRulesEvaluator
+    private val configurationStateManager: ConfigurationStateManager
+    private val configurationRulesManager: ConfigurationRulesManager
+    private val retryWorker: ScheduledExecutorService
+    private val extensionEventListener: ExtensionEventListener = ExtensionEventListener {
```
spoorthipujariadobe, line 71: Nit: Do you think we can pass it inline?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/ConfigurationExtension.kt
```diff
@@ -0,0 +1,454 @@
...
+
+    /**
+     * Responsible for handling the incoming event requests directed towards [ConfigurationExtension].
+     *
+     * @param event the event request dispatched to the [ConfigurationExtension]
+     */
+    internal fun handleConfigurationRequestEvent(event: Event) {
+        when {
+            event.eventData.containsKey(CONFIGURATION_REQUEST_CONTENT_JSON_APP_ID) -> {
+                configureWithAppID(event)
```
spoorthipujariadobe, line 191: Would it be more readable if we used `when`?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/ConfigurationExtension.kt
```diff
@@ -0,0 +1,454 @@
...
+        configurationStateManager.replaceConfiguration(config)
+        publishConfigurationChanges(event, false)
+    }
+
+    /**
+     * Updates the current configuration with the content from [ConfigurationStateManager.CONFIG_BUNDLED_FILE_NAME]
+     *
+     * @param event which contains [CONFIGURATION_REQUEST_CONTENT_JSON_ASSET_FILE] as part of its event data
+     */
+    private fun configureWithFileAsset(event: Event) {
```
spoorthipujariadobe, line 291: Isn't there a public API which also accepts the `fileName`? 

https://aep-sdks.gitbook.io/docs/foundation-extensions/mobile-core/configuration/configuration-api-reference#configurewithfileinassets 

https://git.corp.adobe.com/dms-mobile/bourbon-core-java-core/blob/e6f4e153ef00cb053f699d453e97272ab0db8ca5/code/configuration/src/main/java/com/adobe/marketing/mobile/ConfigurationExtension.java#L312

spoorthipujariadobe: Is fetching the `fileName` from the  `event` missing? Looks like we always use `CONFIG_BUNDLED_FILE_NAME`

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/ConfigurationExtension.kt
```diff
@@ -0,0 +1,454 @@
...
+                TAG,
+                "Invalid configuration. Configuration contains non string keys."
+            )
+            return
+        }
+
+        val programmaticConfig = try {
+            config as? Map<String, Any?>
+        } catch (e: Exception) {
+            null
```
spoorthipujariadobe, line 331: Let's add a log here about invalid config

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/ConfigurationExtension.kt
```diff
@@ -0,0 +1,454 @@
...
+     */
+    private fun dispatchConfigurationResponse(eventData: Map<String, Any?>, triggerEvent: Event?) {
+        val event = Event.Builder(
+            "Configuration Response Event",
+            EVENT_TYPE, EVENT_SOURCE_RESPONSE_CONTENT
+        )
+            .setEventData(eventData)
+            .build()
+
+        MobileCore.dispatchResponseEvent(event, triggerEvent) {
```
spoorthipujariadobe, line 394: Can we use `ExtensionApi.dispatch` instead?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/ConfigurationExtension.kt
```diff
@@ -0,0 +1,454 @@
...
+        }
+    }
+
+    private fun dispatchConfigurationRequest(eventData: Map<String, Any?>) {
+        val event = Event.Builder(
+            "Configure with AppID Internal",
+            EVENT_TYPE, EVENT_SOURCE_REQUEST_CONTENT
+        )
+            .setEventData(eventData).build()
+        MobileCore.dispatchEvent(event) {
```
spoorthipujariadobe, line 405: Same here. Let's use `ExtensionApi.dispatch`

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/ConfigurationExtension.kt
```diff
@@ -0,0 +1,454 @@
...
+        private const val TAG = "ConfigurationExtension"
+        private const val EXTENSION_NAME = "com.adobe.module.configuration"
+        private const val EVENT_TYPE = "com.adobe.eventType.configuration"
+        private const val EVENT_SOURCE_REQUEST_CONTENT = "com.adobe.eventSource.requestContent"
+        private const val EVENT_SOURCE_RESPONSE_CONTENT = "com.adobe.eventSource.responseContent"
+
+        internal const val CONFIGURATION_REQUEST_CONTENT_JSON_APP_ID = "config.appId"
+        internal const val CONFIGURATION_REQUEST_CONTENT_JSON_FILE_PATH = "config.filePath"
+        internal const val CONFIGURATION_REQUEST_CONTENT_JSON_ASSET_FILE = "config.assetFile"
+        internal const val CONFIGURATION_REQUEST_CONTENT_UPDATE_CONFIG = "config.update"
```
praveek, line 48: These constants are already used from our public APIs and are already present in CoreConstants file. Can you remove these and reuse them?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/ConfigurationExtension.kt
```diff
@@ -0,0 +1,454 @@
...
+ * Configuration Extension.
+ * Responsible for retrieving the configuration of the SDK, updating the shared state and
+ * dispatching configuration updates through the `EventHub`
+ */
+internal class ConfigurationExtension : Extension {
+
+    companion object {
+        private const val TAG = "ConfigurationExtension"
+        private const val EXTENSION_NAME = "com.adobe.module.configuration"
+        private const val EVENT_TYPE = "com.adobe.eventType.configuration"
```
praveek, line 41: Use all types and sources from EventType and EventSource classes

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/ConfigurationExtension.kt
praveek: +1. 
I would prefer avoiding this variable and passing it as lambda to `registerEventListener` API.

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/ConfigurationExtension.kt
```diff
@@ -0,0 +1,454 @@
...
+            publishConfigurationChanges(null, true)
+        }
+    }
+
+    override fun getName(): String {
+        return EXTENSION_NAME
+    }
+
+    override fun getFriendlyName(): String {
+        return EXTENSION_NAME
```
praveek, line 180: This should be `Configuration`.

praveek: Can you also override the getVersion() API?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/utility/FileUtils.kt
```diff
@@ -0,0 +1,199 @@
...
+        var entryProcessedSuccessfully = true
+        try {
+            ZipInputStream(FileInputStream(zipFile)).use { zipInputStream ->
+                // get the zipped file list entry
+                var ze = zipInputStream.nextEntry
+                val outputFolderCanonicalPath = folder.canonicalPath
+                if (ze == null) {
+                    // Invalid zip file!
+                    entryProcessedSuccessfully = false
+                    MobileCore.log(LoggingMode.WARNING, TAG, "Zip file was invalid")
```
spoorthipujariadobe, line 157: Nit: Can we fail fast and return `false` here?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/ConfigurationExtension.kt
prudrabhat: You are right! I missed the two public api's for configuration as well as fetching the asset name from the event. Will change accordingly.

code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/utility/FileUtils.kt
prudrabhat: I actually prefer if this class acts as a true utility class that reduces the null check boilerplate code on the caller. That said, I will change it if you feel strongly about making them non null.

code/android-core-library/src/main/java/com/adobe/marketing/mobile/utils/RemoteDownloader.kt
prudrabhat: Agree with eliminating the caching logic and deferring it to the caller of `download(...)`. 
However, based on discussion with @praveek , in the interest of merging Configuration extension, we planned to leave this as is, i.e 1:1 parity with existing production functionality and, then, create a new PR which eliminates caching as a fast follow to this PR. Does it sound good to you?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/utility/FileUtils.kt
yangyansong-adbe: Not having a strong feeling about the above recommendation. We can leave it as is if the callers have to do null checks everywhere.

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/ConfigurationExtension.kt
prudrabhat: The new EventHub, as of now, does not send a boot event when start is called. As the extensions are processing their own queue via the container, I felt that it should be sufficient to move the operations done during boot earlier to onRegistered, similar to the [implementation swift](https://github.com/adobe/aepsdk-core-ios/blob/main/AEPCore/Sources/configuration/Configuration.swift#L44).  Do you see any down side of doing it this way?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/utils/RemoteDownloader.kt
prudrabhat: This will likely be removed once we eliminate caching from RemoteDownloader. I will defer it the fast follow task.

code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/utility/FileUtils.kt
prudrabhat: `.nextEntry` will check and close the current open entry if any. But agree that it is safer to be explicit. Added `closeEntry()` before moving to the next extry.

code/android-core-library/src/main/java/com/adobe/marketing/mobile/configuration/ConfigurationExtension.kt
prudrabhat: Added logs inside ConfigurationStateManager to show what config is being loaded.

prudrabhat:
@yangyansong-adbe , @praveek , @spoorthipujariadobe : 
It looks like my original fork of `aepsdk-core-android` from which this review has been posted was detached when the repo was made public. I addressed the current comments and raised a new PR : https://github.com/adobe/aepsdk-core-android/pull/177 

I also ported the comments that re still pending discussion as comments to the PR. Please continue the review in https://github.com/adobe/aepsdk-core-android/pull/177 . I will mark this as closed now.

PR #146, opened 2022-07-27
Refactoring Signal extension in Kotlin

yangyansong-adbe:


code/android-signal-library/src/main/java/com/adobe/marketing/mobile/signal/EventExtensions.kt
```diff
@@ -0,0 +1,123 @@
...
+
+@JvmSynthetic
+internal fun Event.isCollectPii(): Boolean {
+    return this.consequenceType() == SignalConstants.EventDataKeys.Signal.RULES_RESPONSE_CONSEQUENCE_TYPE_PII
+}
+
+@JvmSynthetic
+internal fun Event.contentType(): String {
+    return try {
+        DataReader.getString(this.details(), SignalConstants.EventDataKeys.RuleEngine.CONTENT_TYPE)
```
praveek, line 36: nit: I would prefer using optString method everywhere

code/android-signal-library/src/main/java/com/adobe/marketing/mobile/signal/SignalExtension.kt
```diff
@@ -0,0 +1,148 @@
...
+    override fun getName(): String {
+        return SignalConstants.EXTENSION_NAME
+    }
+
+    override fun getFriendlyName(): String {
+        return SignalConstants.FRIENDLY_NAME
+    }
+
+    override fun getVersion(): String {
+        return Signal.extensionVersion()
```
praveek, line 50: nit: Can we also move this one to SignalConstants?

code/android-signal-library/src/main/java/com/adobe/marketing/mobile/signal/EventExtensions.kt
yangyansong-adbe: https://github.com/adobe/aepsdk-core-android/blob/09dabcd1b19215f21c68698ffdd9cd779b49d480/code/android-core-library/src/main/java/com/adobe/marketing/mobile/utils/DataReader.java#L135
DataReader throws a runtime exception, so I have to do try-catch here. I can address this comment later after changing the above logic.

code/android-signal-library/src/main/java/com/adobe/marketing/mobile/signal/SignalExtension.kt
yangyansong-adbe: We keep the version constants in the public API class for all extensions, I think we can just keep the consistency for now.

code/android-signal-library/src/androidTest/java/com/adobe/marketing/mobile/signal/EventHubProxy.java
```diff
@@ -8,23 +8,12 @@
...
-	public String getVersion() {
-		return Signal.extensionVersion();
-	}
-
-	public Map<String, String> getAdditionalInfo() {
-		return new HashMap<>();
-	}
-}
\ No newline at end of file
+//package com.adobe.marketing.mobile.signal;
```
praveek, line 11: Are you planning to use this file?

code/android-signal-library/src/main/java/com/adobe/marketing/mobile/signal/Log.kt
```diff
@@ -0,0 +1,37 @@
+/*
```
praveek, line 1: nit: Remove this file and make the Core Log class public as part of this PR.

code/android-signal-library/src/androidTest/java/com/adobe/marketing/mobile/signal/EventHubProxy.java
yangyansong-adbe: Yeah, I will move it to the integration-test module later.

code/android-signal-library/src/main/java/com/adobe/marketing/mobile/signal/SignalExtension.kt
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    private fun handleConfigurationResponse(event: Event?) {
```
praveek, line 53: I added `Nonnull` annotation to `ExtensionEventListener`. We can update all eventListeners to `Event` instead of `Event?`

code/android-signal-library/src/main/java/com/adobe/marketing/mobile/signal/SignalExtension.kt
```diff
@@ -0,0 +1,148 @@
...
+ */
+package com.adobe.marketing.mobile.signal
+
+import com.adobe.marketing.mobile.*
+import com.adobe.marketing.mobile.services.HitQueuing
+import com.adobe.marketing.mobile.services.PersistentHitQueue
+import com.adobe.marketing.mobile.services.ServiceProvider
+import com.adobe.marketing.mobile.utils.DataReader
+
+class SignalExtension(extensionApi: ExtensionApi?) : Extension(extensionApi) {
```
spoorthipujariadobe, line 19: Any reason why `ExtensionApi` is an optional?

code/android-signal-library/src/main/java/com/adobe/marketing/mobile/signal/SignalHitProcessor.kt
```diff
@@ -0,0 +1,117 @@
...
+ */
+package com.adobe.marketing.mobile.signal
+
+import androidx.annotation.VisibleForTesting
+import com.adobe.marketing.mobile.services.*
+import java.util.concurrent.CountDownLatch
+import java.util.concurrent.TimeUnit
+
+internal class SignalHitProcessor : HitProcessing {
+    private val networkService: Networking
```
spoorthipujariadobe, line 19: We allow overriding network service. Let's make it an optional and add null checks till we find a better way of handling such use cases.

code/android-signal-library/src/main/java/com/adobe/marketing/mobile/signal/SignalHitProcessor.kt
```diff
@@ -0,0 +1,117 @@
...
+        private const val LOG_TAG = "SignalHitProcessor"
+        private const val HIT_QUEUE_RETRY_TIME_SECONDS = 30
+    }
+
+    constructor() {
+        this.networkService = ServiceProvider.getInstance().networkService
+    }
+
+    @VisibleForTesting
+    internal constructor(networkService: Networking) {
```
spoorthipujariadobe, line 31: Maybe we can set it directly in ServiceProvider instead of having a constructor for it?

code/android-signal-library/src/main/java/com/adobe/marketing/mobile/signal/SignalExtension.kt
yangyansong-adbe: Will do.

code/android-signal-library/src/main/java/com/adobe/marketing/mobile/signal/SignalExtension.kt
yangyansong-adbe: https://github.com/adobe/aepsdk-core-android/blob/258d31272a6dd41f3fdacd2c6045297f93db6a64/code/android-core-library/src/main/java/com/adobe/marketing/mobile/Extension.java#L28
As we discussed before, we will not add `NotNull` annotation to `Extension` class for backward compatibility. So, technically the Kotlin constructor could take in null.

code/android-signal-library/src/main/java/com/adobe/marketing/mobile/signal/SignalHitProcessor.kt
yangyansong-adbe: https://github.com/adobe/aepsdk-core-android/blob/258d31272a6dd41f3fdacd2c6045297f93db6a64/code/android-core-library/src/phone/java/com/adobe/marketing/mobile/services/ServiceProvider.java#L124
The `ServiceProvide` has done the null check

code/android-signal-library/src/main/java/com/adobe/marketing/mobile/signal/SignalHitProcessor.kt
yangyansong-adbe: We can't do that as `ServiceProvider` is in Core, another package.

code/android-signal-library/src/main/java/com/adobe/marketing/mobile/signal/SignalExtension.kt
spoorthipujariadobe: Considering the constructor is `protected` is there a possibility of getting a `null` value? @prudrabhat might need to do the same in configuration as well

spoorthipujariadobe (approved):

code/android-signal-library/src/main/java/com/adobe/marketing/mobile/signal/SignalExtension.kt
praveek: This won't be non null. I have also added `NonNull` annotation to Extension class to force kotlin implementations to use non null type.

praveek (approved):

PR #161, opened 2022-08-05
Add MobileCore.registerExtensions API

praveek:
https://github.com/adobe/aepsdk-core-android/issues/155 - Added `MobileCore.registerExtensions` API to simplify extension registration

```
public static void registerExtensions(@NonNull final List<Class<? extends Extension>> extensions, final AdobeCallback<?> completionCallback)
```

Currently, all our product extensions expose a single public API class with static methods. Registration with EventHub is abstracted by `PRODUCT.registerExtension()` method and the classes which implement extension APIs are not public.

We need to make them public for apps to use registerExtensions API. We can follow one of these two patterns. 

1)  Have a single public class which extends `Extension` class. This class should also expose current public APIs as static methods. (Sample1, Sample1Kt)
```
List<Class<? extends Extension>> extensions = Arrays.asList(Edge.class, Identity.class);
MobileCore.registerExtensions(extensions, new AdobeCallback() {
	@Override
	public void call(Object value) {
                Edge.sendEvent(...);                
                Identity.getUrlVariables(...);
	}
});
```

2) Make the extension classes public. For each product, it should be PRODUCT name with Extension suffix. (Sample2, Sample2Kt)
```
List<Class<? extends Extension>> extensions = Arrays.asList(EdgeExtension.class, IdentityExtension.class);
MobileCore.registerExtensions(extensions, new AdobeCallback() {
	@Override
	public void call(Object value) {
                Edge.sendEvent(...);                
                Identity.getUrlVariables(...);
	}
});
```

The PR contains sample extensions both in Java and Kotlin. Both these can be implemented without breaking current apps and i recommend all our extensions to follow a similar pattern for consistency.


code/android-core-library/src/phone/java/com/adobe/marketing/mobile/MobileCore.java
```diff
@@ -200,6 +203,39 @@ public static void log(@NonNull final LoggingMode mode, final String tag, final
         }
     }
 
+    public static void registerExtensions(@NonNull final List<Class<? extends Extension>> extensions, final AdobeCallback<?> completionCallback) {
```
yangyansong-adbe, line 206: How about using `variable arguments` (varargs) here?  For example: https://stackoverflow.com/questions/26170318/varargs-of-type-class-in-java

praveek: I don't prefer using `variable arguments` as they are internally translated into arrays. It makes it easier to incorrectly use our API as java only allows creating array of unbounded wildcards.

yangyansong-adbe: I don't think it's an issue as it's quite common exposing `varargs` arguments in Java, at the same time, the `varargs` provide a more friendly API style:
```java
public static void registerExtensions(final AdobeCallback<?> completionCallback, Class<? extends Extension>... extensions)
```
```java
MobileCore.registerExtensions(
                 null,
                Signal.class,
                Lifecycle.class
                );
```

praveek: I feel, the API usage will be difficult for the use case where customers have to optionally include extensions. 
if (useLifecycleExtension) {
   MobileCore.registerExtensions(callback, Signal.class, Lifecycle.class);
} else {
   MobileCore.registerExtensions(callback, Signal.class);
}

or they will have to use arrays which could lead to type safety issues. 
Class<? extends Extension>[] extensionList = new Class[] { ... , ... };
MobileCore.registerExtensions(callback, extensionList)

yangyansong-adbe: https://github.com/adobe/aepsdk-core-ios/blob/1f037918856101e863a8e6b1bc44e04ca40f189f/AEPCore/Sources/core/MobileCore.swift#L85
For your use case, I think adding another API like the above Swift API that accepts one extension class should work.

code/android-core-library/src/phone/java/com/adobe/marketing/mobile/MobileCore.java
```diff
@@ -200,6 +203,39 @@ public static void log(@NonNull final LoggingMode mode, final String tag, final
...
+        if(!sdkInitializedWithContext.get()) {
+            Log.error(LOG_TAG, "Failed to registerExtension - setApplication not called");
```
pfransenadb, line 208: nit: should be "Failed to registerExtensions"

code/android-core-library/src/phone/java/com/adobe/marketing/mobile/MobileCore.java
```diff
@@ -200,6 +203,39 @@ public static void log(@NonNull final LoggingMode mode, final String tag, final
...
+            return;
+        }
+
+        List<Class<? extends Extension>> allExtensions = new ArrayList<>();
+        // allExtensions.add(ConfigurationExtension.class);
```
pfransenadb, line 213: remove comment

code/android-core-library/src/phone/java/com/adobe/marketing/mobile/MobileCore.java
```diff
@@ -200,6 +203,39 @@ public static void log(@NonNull final LoggingMode mode, final String tag, final
...
+        if (extensions != null) {
+            for(final Class<? extends Extension> extension: extensions) {
+                if (extension != null) {
+                    allExtensions.add(extension);
+                }
+            }
+        }
+
+        AtomicInteger registeredExtensions = new AtomicInteger(0);
```
pfransenadb, line 222: should be final

code/android-core-library/src/phone/java/com/adobe/marketing/mobile/MobileCore.java
```diff
@@ -200,6 +203,39 @@ public static void log(@NonNull final LoggingMode mode, final String tag, final
...
+        for (final Class<? extends Extension> extension: allExtensions) {
+            EventHub.Companion.getShared().registerExtension(extension, eventHubError -> {
+                Log.debug(LOG_TAG, "Finished registering extension " + extension + "with status "+ eventHubError);
```
pfransenadb, line 225: nit: get rid of "Finished" here, it's confusing with the log statement below.  "Registered extension" would be better.

code/android-core-library/src/phone/java/com/adobe/marketing/mobile/MobileCore.java
pfransenadb, line 206: What happens if:

a. this function is called more than once?  (at face value the registeredExtensions atomic int is going to cause some problems because it's initialized to zero)

b. if this function is called more than once across threads? (less important but still potentially a problem)

pfransenadb (changes requested):
Headed in the right direction, there's some potential issues with registerExtensions that need to be addressed (and a few nits around comments, log statements, and final variables).

code/android-core-library/src/phone/java/com/adobe/marketing/mobile/MobileCore.java
praveek: @pfransenadb This function behaves similarly to the one we have in Swift SDK. It registers the extensions passed to it with the event hub, starts event processing and then notifies the caller by calling `completionCallback`. 

We expect customers to simplify SDK initialization by calling new registerExtensions() API once although calling it multiple times will work properly. 

a. Each call to `registerExtensions` will call the `completionCallback` after registering the extensions passed to it with EventHub. `registeredExtensions` atomic int is a local variable and only tracks extensions passed to a particular call. 

b. EventHub is thread safe and processes everything in its own queue. Calling from different threads should work as expected.

pfransenadb (approved):
👍

Issue #150, opened 2022-07-28
Remove MobileCore.getEventHistory API


praveek:
- Remove MobileCore.getEventHistory public API 
- Add an API to query EventHistory.

PR #162, opened 2022-08-08
Add a `getHistoricalEvents` API (port Swift changes)

yangyansong-adbe:
Exposes a `getHistoricalEvents` API in ExtensionApi class to follow the Swift implementation.
## Related Issue
-  #150


praveek (changes requested):
Can you also update the historical condition to query `from extensionAPI.getHistoricalEvents` API?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/eventhub/EventHub.kt
```diff
@@ -182,7 +198,15 @@ internal class EventHub {
             )
         }
 
-        // TODO: Record event to event history database if required.
+        event.mask?.let {
```
praveek, line 201: This should be moved to SerialWorkDispatcher.WorkHandler. We already have a Todo marked there for this logic. 
By keeping the logic here, we will miss recording events generated as result of dispatch consequence.

code/android-core-library/src/main/java/com/adobe/marketing/mobile/launch/rulesengine/json/JSONCondition.kt
```diff
@@ -38,30 +39,34 @@ internal abstract class JSONCondition {
...
+                            extensionApi
                         )
                     )
                     TYPE_VALUE_MATCHER -> MatcherCondition(
                         JSONDefinition.buildDefinitionFromJSON(
                             jsonCondition.getJSONObject(
                                 KEY_DEFINITION
-                            )
+                            ),
+                            extensionApi
```
spoorthipujariadobe, line 59: Has `MatcherCondition` and `GroupCondition` been changed to accept `extensionApi`? Don't see it in this PR

spoorthipujariadobe (approved):

praveek (approved):

code/android-core-library/src/main/java/com/adobe/marketing/mobile/launch/rulesengine/json/JSONCondition.kt
yangyansong-adbe: No, we don't need to add it to other conditions as they don't depend on `ExtensionApi`.

PR #163, opened 2022-08-08
[quick-fix] fix the use of the wrong API

yangyansong-adbe:


code/android-core-library/src/phone/java/com/adobe/marketing/mobile/MobileCore.java
```diff
@@ -89,7 +89,7 @@ public static void setApplication(@NonNull final Application application) {
             return;
         }
 
-        if (sdkInitializedWithContext.compareAndSet(false, true)) {
+        if (sdkInitializedWithContext.getAndSet(true)) {
```
praveek, line 92: I missed the `!` part in the if check but this should also work for our use case.

praveek (approved):

PR #181, opened 2022-08-23
Add a public logging service and make Log class public as a utility class

yangyansong-adbe:


code/android-core-library/src/main/java/com/adobe/marketing/mobile/Log.java
```diff
@@ -10,135 +10,123 @@
...
  */
-class Log {
-	private static LoggingService loggingService = null;
-	private static LoggingMode loggingMode = LoggingMode.ERROR;
-
-	static final String UNEXPECTED_NULL_VALUE = "Unexpected Null Value";
-	static final String UNEXPECTED_EMPTY_VALUE = "Unexpected Empty Value";
-	static final String INVALID_FORMAT = "Invalid Format";
+public class Log {
+    private static Logging loggingService = ServiceProvider.getInstance().getLoggingService();
```
praveek, line 23: This will work for now. We should get the instance just before logging when we add a method to override Logging service.

code/android-core-library/src/phone/java/com/adobe/marketing/mobile/AndroidLoggingService.java
```diff
@@ -0,0 +1,32 @@
+package com.adobe.marketing.mobile;
```
praveek, line 1: nit: We can remove this class. I don't think any extension outside Core is using this.

praveek (approved):
Can you create a GH issue to add an API to override log service?

PR #183, opened 2022-08-23
Move EventHistoryRequest and EventHistoryResultHandler to the main package

yangyansong-adbe:


spoorthipujariadobe (approved):

code/android-core-library/src/main/java/com/adobe/marketing/mobile/EventHistoryRequest.java
```diff
@@ -0,0 +1,53 @@
...
+        this.map = map;
+        this.fromDate = fromDate;
+        this.toDate = toDate;
+    }
+
+    public long getMaskAsDecimalHash() {
+        StringBuilder kvpStringBuilder = new StringBuilder();
+        if (map != null && !map.isEmpty()) {
+            for (Map.Entry<String, Object> entry : map.entrySet()) {
+                kvpStringBuilder.append(entry.getKey() + ":" + entry.getValue());
```
praveek, line 40: Can we reuse this Util function so that we don't duplicate the logic? 
`return MapUtilsKt.convertMapToFnv1aHash(map, null);`

praveek (approved):

PR #186, opened 2022-08-26
Add a new parameter to Log APIs

yangyansong-adbe:


code/android-core-library/src/main/java/com/adobe/marketing/mobile/Log.java
```diff
@@ -57,77 +57,102 @@ static LoggingMode getLogLevel() {
...
+     * @param format    the string format to be logged
+     * @param params    values to be inserted into the format
      * @see LoggingMode
      */
-    public static void debug(final String source, final String format, final Object... params) {
+    public static void debug(final String extension, final String source, final String format, final Object... params) {
         if (loggingService != null && loggingMode.id >= LoggingMode.DEBUG.id) {
             try {
-                loggingService.debug(source, String.format(format, params));
+                loggingService.debug(extension + "-" + source, String.format(format, params));
```
praveek, line 95: Can you do a null check for extension to avoid printing `null`?

yangyansong-adbe: Ops, it will print the `null` String. Ok, how about using an empty String instead?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/Log.java
```diff
@@ -13,6 +13,8 @@
 import com.adobe.marketing.mobile.services.Logging;
 import com.adobe.marketing.mobile.services.ServiceProvider;
 
+import org.jetbrains.annotations.NotNull;
```
praveek, line 16: Wrong import

praveek (approved):

PR #189, opened 2022-08-31
Create a new public Log class as the public Logging utility

yangyansong-adbe:


code/testapp-kotlin/src/main/java/com/adobe/marketing/mobile/app/kotlin/CoreView.kt
```diff
@@ -107,6 +108,11 @@ fun CoreView(navController: NavHostController) {
             }) {
                 Text(text = "getLogLevel")
             }
+            Button(onClick = {
+                Log.debug("source", "%s loginggggg", "x")
```
praveek, line 112: `logging`

code/android-core-library/src/phone/java/com/adobe/marketing/mobile/MobileCore.java
```diff
@@ -165,7 +165,7 @@ public static void setLogLevel(@NonNull LoggingMode mode) {
      */
     @NonNull
     public static LoggingMode getLogLevel() {
-        return Log.getLogLevel();
+        return com.adobe.marketing.mobile.services.Log.getLogLevel();
```
yangyansong-adbe, line 168: @praveek  we can clean up the core extension after the legacy classes are moved to a separate artifact.( then we won't see 2 Log classes in core extension)

praveek (approved):

PR #198, opened 2022-09-08
Add integration tests for Signal

yangyansong-adbe:


code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/configuration/ConfigurationExtension.kt
```diff
@@ -489,16 +489,10 @@ internal class ConfigurationExtension : Extension {
...
-        val builder = Event.Builder(
+        val event = Event.Builder(
             "Configuration Response Event",
             EventType.CONFIGURATION, EventSource.RESPONSE_CONTENT
-        ).setEventData(eventData)
-
-        val event: Event = if (triggerEvent == null) {
-            builder.build()
-        } else {
-            builder.inResponseToEvent(triggerEvent).build()
```
prudrabhat, line 500: `dispatchConfigurationResponse(..)`  should always honor the trigger event that is passed.
I think the fix here to maintain [parity with swift](https://github.com/adobe/aepsdk-core-ios/blob/55d208283eed4f99b5fe8ed75940fcc03b0511a1/AEPCore/Sources/configuration/Configuration.swift#L229) should be to do
`dispatchConfigurationResponse(config, null)` instead of passing the trigger event [here](https://github.com/adobe/aepsdk-core-android/blob/dev-v2.0.0/code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/configuration/ConfigurationExtension.kt#L477)

code/integration-tests/src/androidTest/java/com/adobe/marketing/mobile/integration/signal/SignalExtensionIntegrationTests.kt
```diff
@@ -0,0 +1,223 @@
...
+        )
+        MobileCore.dispatchEvent(
+            Event.Builder("Test", "type", "source")
+                .setEventData(
+                    mapOf(
+                        "name" to "testGetRequest"
+                    )
+                ).build()
+        )
+        countDownLatch.await(10, TimeUnit.MILLISECONDS)
```
spoorthipujariadobe, line 169: nit: could we assertFalse here to ensure the thread wasn't interrupted before wait time elapsed?

spoorthipujariadobe (approved):

code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/configuration/ConfigurationExtension.kt
yangyansong-adbe: Discussed it offline, and we will make some changes in another PR.

yangyansong-adbe:
Merge this PR for now, the configuration change is only a quick fix to unblock the integration tests. Also offline discussed with @prudrabhat, we will create another PR to clean up the configuration logic and add more tests.

PR #220, opened 2022-09-22
Add public utils and cleanup existing dependencies 

praveek:
Added few public utils and fixed imports.


spoorthipujariadobe (approved):

code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/util/FileUtils.kt
```diff
@@ -21,7 +21,7 @@ import java.io.InputStream
 import java.io.InputStreamReader
 import java.util.zip.ZipInputStream
 
-object FileUtils {
+internal object FileUtils {
```
prudrabhat, line 24: This utility has usages outside of the `internal` package with my changes for `CacheService`. What do you think about moving this to public utils?

praveek: @prudrabhat Let us review and make it public after your changes are in. I prefer exposing only `extractFromZip` unless other extensions use any of the other methods.

prudrabhat: Sure. 👍🏼  `readAsString()` and `extractFromZip` are the ones with usage in rule download.

prudrabhat (approved):

PR #224, opened 2022-09-26
Fix issues in RulesEngine and EventHub

praveek:
Fix issues in RulesEngine public APIs
Add missing tests and fix inconsistencies in EventHub and RulesEngine.


code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/eventhub/ExtensionContainer.kt
```diff
@@ -299,6 +300,7 @@ internal class ExtensionContainer constructor(
         )
     }
 
+    @Deprecated("Deprecated in Java")
```
prudrabhat, line 303: nit: `Deprecated in Java` may be a bit confusing for implementers using other JVM languages. Suggest switching to `Deprecated`

prudrabhat (approved):

PR #228, opened 2022-09-28
Bug fix to reprocess events when `readyToProcess` returns false

praveek:
Bug fix to reprocess events when `readyToProcess` returns false
Tests to verify `getSharedState` APIs return correct state with barrier. 
Made URLBuilder utility class public


code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/SerialWorkDispatcher.kt
```diff
@@ -66,18 +66,22 @@ open class SerialWorkDispatcher<T>(private val name: String, private val workHan
     /**
      * Represents the functional interface that is responsible for doing the desired work on each item of the [workQueue].
      * [WorkHandler.doWork] is called from the background worker thread that the [SerialWorkDispatcher] maintains.
+     * If [WorkHandler.doWork] returns false after processing an item, the [SerialWorkDispatcher] stops processing the work queue without removing the item. It restarts when
```
prudrabhat, line 69: Suggest moving these three lines to the `doWork()` doc as it describes the expectations of implementing that method.

prudrabhat (approved):
LGTM. Minor doc comment.

PR #232, opened 2022-10-03
Support 1.x MobileCore initialization

praveek:
This PR contains changes to support SDK initialization for customers migrating from 1.x SDK with no compilation errors or runtime inconsistencies. 

1.x initialization code
```
   try {
         Analytics.registerExtension();
         Edge.registerExtension();
         Identity.registerExtension();
         Lifecycle.registerExtension();
         Signal.registerExtension();
         MobileCore.start(new AdobeCallback () {
                @Override
                public void call(Object o) {
                    MobileCore.configureWithAppID("launch-EN5640fb1a17ec4f8685142ebb71e9ca3c-development");
                }
         });
   } catch (InvalidInitException ex) {
         ...
   }
```

Background: 
    Core 2.0 added [`MobileCore.registerExtensions`](https://github.com/adobe/aepsdk-core-android/pull/161) API to simplify SDK initialization. EventHub currently initializes extensions in background thread and the caller has to call `EventHub.start()` only after registering all extensions. 

Changes: 

MobileCore 
- The SDK will start processing events after it sees the first `MobileCore.registerExtensions` or `MobileCore.start` call.
- After starting event processing
     - Calls to `MobileCore.start()` will be ignored. 
     - Calls to `MobileCore.registerExtension` or `MobileCore.registerExtensions()` will register extensions and update the eventhub shared state. 

EventHub 
- `EventHub.start()` will start processing events only after all `registerExtension` calls before it completes. 
- `EventHub.registerExtension()` will shutdown the ExtensionContainer in case of failed registration. 
- Dispatch callbacks and response listeners from different thread so that it does not block event hub and event processing. 
- Made `InvaidInitException` an unchecked exception to avoid compilation errors.

### DIFF
--- a/code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/eventhub/EventHub.kt
+++ b/code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/eventhub/EventHub.kt
@@ -216,8 +216,14 @@ internal class EventHub(val eventHistory: EventHistory?) {
     /**
      * `EventHub` will begin processing `Event`s when this API is invoked.
      */
+    @JvmOverloads
     fun start(completion: (() -> Unit)? = null) {
         eventHubExecutor.submit {
+            if (hubStartReceived) {
+                Log.debug(CoreConstants.LOG_TAG, LOG_TAG, "Dropping start call as it was already received")
+                return@submit
+            }
+
             this.hubStartReceived = true
             this.hubStartCallback = completion
 
--- a/code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/eventhub/EventHub.kt
+++ b/code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/eventhub/EventHub.kt
@@ -184,7 +184,6 @@ internal class EventHub(val eventHistory: EventHistory?) {
 
     init {
         registerExtension(EventHubPlaceholderExtension::class.java)
-        registerExtension(ConfigurationExtension::class.java)
     }
 
     private var _wrapperType = WrapperType.NONE
@@ -299,6 +298,7 @@ internal class EventHub(val eventHistory: EventHistory?) {
      * @param extensionClass The class of extension to register
      * @param completion Invoked when the extension has been registered or failed to register
      */
+    @JvmOverloads
     fun registerExtension(
         extensionClass: Class<out Extension>,
         completion: ((error: EventHubError) -> Unit)? = null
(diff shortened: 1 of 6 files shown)


pfransenadb (approved):
Makes sense to me.
👍

code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/eventhub/EventHub.kt
```diff
@@ -173,15 +214,45 @@ internal class EventHub(val eventHistory: EventHistory?) {
...
+            }
+
+            this.hubStartReceived = true
+            this.hubStartCallback = completion
+
+            tryStartHub()
         }
     }
 
+    private fun tryStartHub() {
```
prudrabhat, line 232: This method is being called from 2 different places 
- `start()`  invoked on `eventHubExecutor` thread
- `extensionPostRegistration()` invoked on `extensionInitExecutor` thread.

If both the invocations execute simultaneously (say for the last extension registration followed by mobile core start in customer app code) then we may end up in a situation where the hub is never started.
Wonder if we should  make the boolean gates to be volatile or serialize these operations here.

praveek: `extensionPostRegistration` is also invoked on `eventHubExecutor`.
https://github.com/praveek/aepsdk-core-android/blob/dev-v2.0.0/code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/eventhub/EventHub.kt#L316

code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/eventhub/EventHub.kt
```diff
@@ -226,50 +297,92 @@ internal class EventHub(val eventHistory: EventHistory?) {
...
+
+        if (error != EventHubError.None) {
+            unregisterExtensionInternal(extensionClass)
+        } else {
+            shareEventHubSharedState()
+        }
+
+        if (!hubStarted) {
+            registrationRequestsBeforeStart.remove(extensionClass)
+            tryStartHub()
```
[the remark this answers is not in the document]
praveek: `shareEventHubSharedState` checks if the event hub has started before updating the shared state. 
This method is called in `extensionPostRegistration` to update the shared state when registering extensions after starting the event hub.

code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/eventhub/EventHub.kt
prudrabhat: Ah! My bad. Misread it. LGTM.

prudrabhat (approved):

PR #241, opened 2022-10-07
Add a method for testing - MobileCore.resetSDK()

yangyansong-adbe:


praveek:
I don't recommend adding a public method to MobileCore. Can you change it to default accessor? Tests can access it by adding a helper class in com.adobe.marketing.mobile package.

yangyansong-adbe:
https://developer.android.com/reference/androidx/annotation/VisibleForTesting
@praveek  It's only a public method in test code.

code/android-core-library/src/phone/java/com/adobe/marketing/mobile/MobileCore.java
```diff
@@ -991,4 +991,10 @@ public static boolean dispatchResponseEvent(@NonNull final Event responseEvent,
         return true;
     }
 
+    @VisibleForTesting
+    public static void resetSDK() {
```
prudrabhat, line 995: A public method visible for testing is a bit odd for an implementer using the SDK. What do you think about adding a test utility helper in the same package?

yangyansong-adbe:
@praveek @prudrabhat addressed your review comments.

praveek (approved):

prudrabhat (approved):

code/android-core-library/src/phone/java/com/adobe/marketing/mobile/MobileCore.java
```diff
@@ -991,4 +992,10 @@ public static boolean dispatchResponseEvent(@NonNull final Event responseEvent,
         return true;
     }
 
+    @VisibleForTesting
+    static void resetSDK() {
+        EventHub.Companion.setShared(new EventHub());
```
praveek, line 997: Can you also shutdown the old event hub instance before setting a new one?

praveek (changes requested):
Shutdown the event hub instance to avoid race conditions in test.

praveek (approved):

PR #244, opened 2022-10-19
change HitProcessing.processHit to an async API

yangyansong-adbe:


code/android-signal-library/src/main/java/com/adobe/marketing/mobile/signal/SignalHitProcessor.kt
```diff
@@ -42,18 +36,19 @@ internal class SignalHitProcessor : HitProcessing {
         return HIT_QUEUE_RETRY_TIME_SECONDS
     }
 
-    override fun processHit(entity: DataEntity?): Boolean {
-        if (entity == null) {
-            Log.warning(SignalConstants.LOG_TAG, CLASS_NAME,"Drop this data entity as it is null.")
-            return true
+    override fun processHit(entity: DataEntity, processingResult: HitProcessingResult) {
+        if (entity == null || processingResult == null) {
+            Log.warning(SignalConstants.LOG_TAG, CLASS_NAME, "Drop this data entity as it is null.")
```
praveek, line 41: You missed calling completion handler here

code/android-signal-library/src/main/java/com/adobe/marketing/mobile/signal/SignalHitProcessor.kt
```diff
@@ -92,7 +87,7 @@ internal class SignalHitProcessor : HitProcessing {
             countDownLatch.countDown()
         }
         countDownLatch.await((request.connectTimeout + 1).toLong(), TimeUnit.SECONDS)
-        return result
+        processingResult.complete(result)
```
praveek, line 90: Can you avoid the latch and call `processingResult.complete(result)` inside the http handler?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/services/HitProcessing.java
```diff
@@ -11,21 +11,25 @@
...
-	 * @return Hit retry interval in seconds.
-	 */
-	int retryInterval(DataEntity entity);
+    /**
+     * Determines the interval at which a hit should be retried
+     *
+     * @param entity The hit whose retry interval is to be computed
+     * @return Hit retry interval in seconds.
+     */
+    int retryInterval(DataEntity entity);
```
praveek, line 25: Can you add the `@Nonnull` annotation here too?

code/android-signal-library/src/main/java/com/adobe/marketing/mobile/signal/SignalHitProcessor.kt
yangyansong-adbe: `processingResult` could be null here, we shouldb't call the completion handler. And I don't think we need to add more logic to do it when the `entity` is null, and the `processingResult` is not null. 
It's a sanity check, the code should never get there since the arguments should be @NonNull.

praveek: We can remove these null checks here, as the SDK always sends a non-null param. You can also annotate these params `@Nonnull`. It is confusing when reading code as returning without calling completion handler blocks the entire Hit processor.

praveek (approved):

Issue #248, opened 2022-10-26
Add functionality to set custom headers to RulesLoader.loadFromUrl()


rymorale:

praveek:
Reopen if we have a requirement for this API

PR #249, opened 2022-10-26
Add functionality to set custom headers when calling RulesLoader.loadFromUrl

rymorale:
- Add functionality to set custom headers when calling RulesLoader.loadFromUrl

## Related Issue
https://github.com/adobe/aepsdk-core-android/issues/248

## Types of changes

## Checklist:

### DIFF
--- a/code/android-core-library/src/main/java/com/adobe/marketing/mobile/launch/rulesengine/download/RulesLoader.java
+++ b/code/android-core-library/src/main/java/com/adobe/marketing/mobile/launch/rulesengine/download/RulesLoader.java
@@ -77,19 +77,6 @@ public class RulesLoader {
 
     /**
      * Loads rules from the {@code url} and invokes {@code callback} with the extracted rules.
-     * Additionally, the extracted content is cached in cache bucket with name {@code name} and {@code url} as the key
-     * in {@code CacheService}.
-     *
-     * @param url      the url from which the compressed rules are to be downloaded
-     * @param callback the callback that will be invoked with the result of the download
-     */
-    public void loadFromUrl(@NonNull final String url,
-                            @NonNull final AdobeCallback<RulesLoadResult> callback) {
-        loadFromUrl(url, null, callback);
-    }
-
-    /**
-     * Loads rules from the {@code url} and invokes {@code callback} with the extracted rules.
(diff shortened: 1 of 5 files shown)


code/android-core-library/src/main/java/com/adobe/marketing/mobile/launch/rulesengine/download/RulesLoader.java
```diff
@@ -88,19 +85,41 @@ public RulesLoader(@NonNull final String cacheName) {
...
+     * Loads rules from the {@code url} and invokes {@code callback} with the extracted rules.
+     * Any provided {@code customHeaders} are merged with the extracted cache headers and used in the rules download request.
+     * Additionally, the extracted content is cached in cache bucket with name {@code name} and {@code url} as the key
+     * in {@code CacheService}.
+     *
+     * @param url               the url from which the compressed rules are to be downloaded
+     * @param customHeaders     custom headers to include in the rules download request
+     * @param callback          the callback that will be invoked with the result of the download
+     */
+    public void loadFromUrl(@NonNull final String url, @Nullable final Map<String, String> customHeaders,
```
prudrabhat, line 101: Basing this comment on my understanding that RulesLoader is currently being used by Configuration, Campaign and, that the current core release is still in beta. 
I am wondering if it may be better to modify the existing public api itself (instead of adding a new cascading call). I understand that this means we are breaking a _public api_ but, in my view that would make the api surface cleaner for this public class before we move out of beta release -- especially since this api usage within core/configuration can be modified without any external breaking changes and, the api usage is being currently worked on in Campaign extension. What are your thoughts?
I would also like to know what @praveek  feels about this.

rymorale: 👍 i think your point is a good one as core is still in prerelease.

praveek: I agree, we can change the existing API. Are you planning to make a similar change in Swift?

praveek:
Closing this as we no longer need this API

PR #266, opened 2022-11-07
Fix the Identity shared state issue

yangyansong-adbe:


code/android-identity-library/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
```diff
@@ -773,7 +773,11 @@ boolean handleSyncIdentifiers(final Event event, final ConfigurationSharedStateI
 
         // Update share state and persistence here. Any state changes from the server response will be included in the next shared state
         // it's more important to not block other extensions with an IdentityExtension pending state
-        getApi().createSharedState(packageEventData(), event);
+        if (forceSync) {
```
praveek, line 776: I see you dispatch a new event to handle sync event. This check is no longer required and should always pass the event as parameter when creating shared state.

yangyansong-adbe: I intended to add it to handle the Identity initialization logic here: https://github.com/adobe/aepsdk-core-android/pull/266/files#diff-96bcd64970dfe00717733ab220b0fb3b8994e8cb21d42f9bc549c8c365e53ad7R182. The default/first `force sync` should happen before handling other Identity requests, which means it has to be a sync call to block the Identity event processing, so we can't dispatch an event for it and have to do the above specific handling.

praveek (approved):

Issue #231, opened 2022-09-29
Add functionality to provide initialization job for SerialWorkDispatcher.


praveek:
SerialWorkDispatcher maintains its own executor to process events. Add an API to set an initialization job which runs in the same executor. ExtensionContainer can use this functionality to initialize the extension.

PR #267, opened 2022-11-07
Add functionality to set initial and final jobs to SerialWorkDispatcher 

prudrabhat:
## Description
- SerialWorkDispatcher maintains its own executor to process events. Adding an API to set an initialization job which runs in the same executor will allow components like ExtensionContainer to use this functionality to initialize the extension.
This will allow getting rid of one cached thread pool executor.
- Remove the protected apis prepare() and cleanup() as components can use `setInitalJob` and `setFinalJob` instead. This is a breaking change to SerialDispatcher but felt like this is OK given that we are in pre release.
- Add tests

## Related Issue

- #231 

## Motivation and Context
SerialWorkDispatcher maintains its own executor to process events. Adding an API to set an initialization job which runs in the same executor will allow components like ExtensionContainer to use this functionality to initialize the extension.
This will allow  getting rid of one cached thread pool executor.

## How Has This Been Tested?
- Unit tests

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/eventhub/ExtensionContainer.kt
```diff
@@ -91,56 +89,60 @@ internal class ExtensionContainer constructor(
...
-                CoreConstants.LOG_TAG,
-                getTag(),
-                "ExtensionContainer started processing events"
-            )
-        }
+    init {
+
+        eventProcessor.setInitialJob(initJob)
+        eventProcessor.setFinalJob(teardownJob)
+        // Start event processor now as extensions can add event listeners onRegistered() callback
```
praveek, line 134: Can you remove this coment?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/SerialWorkDispatcher.kt
```diff
@@ -144,10 +182,9 @@ open class SerialWorkDispatcher<T>(private val name: String, private val workHan
      * [SerialWorkDispatcher]) before processing starts.
      * Invoked on the thread that calls [start]
```
prudrabhat, line 183: This javadoc is incorrect. Will update it.

praveek (approved):

PR #268, opened 2022-11-08
Cleanup logs and hide unwanted APIs

praveek:
Cleanup logs
Cleanup public APIs
Update Makefile 
Remove unwanted tests


code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/eventhub/EventHub.kt
```diff
@@ -160,6 +160,14 @@ internal class EventHub(val eventHistory: EventHistory?) {
                 it.eventProcessor.offer(processedEvent)
             }
 
+            if (Log.getLogLevel() >= LoggingMode.DEBUG) {
```
yangyansong-adbe, line 163: Any reason that makes us check the log level here?

yangyansong-adbe (approved):

code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/eventhub/EventHub.kt
praveek: The log message converts the entire event to string. This then gets dropped based on the log level. 
I added the check to avoid the string conversion in the hot path.

PR #269, opened 2022-11-08
Add more performance tests

yangyansong-adbe:


code/testapp-kotlin/build.gradle
```diff
@@ -51,22 +51,22 @@ android {
...
     implementation project(path: ':android-core-library')
     implementation project(path: ':android-identity-library')
 
     testImplementation 'junit:junit:4.13.2'
     androidTestImplementation 'androidx.test.ext:junit:1.1.3'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.4.0'
-    androidTestImplementation "androidx.compose.ui:ui-test-junit4:$compose_version"
-    debugImplementation "androidx.compose.ui:ui-tooling:$compose_version"
-    debugImplementation "androidx.compose.ui:ui-test-manifest:$compose_version"
+//    androidTestImplementation "androidx.compose.ui:ui-test-junit4:$compose_version"
```
praveek, line 69: Can you remove them if they are not needed?

praveek:
Can you also add a make target to easily run these tests?

yangyansong-adbe:
> Can you also add a make target to easily run these tests?

@praveek  
Most possibly, we only trigger the test run in Android Studio with a real device connected. So, I think we don't need to add a make target for it.

praveek (approved):

PR #286, opened 2022-12-01
Update TimeUtil helper method for ISO 8601 UTC conversion

timkimadobe:
## Description

This PR updates the helper method `getIso8601DateTimeZoneISO8601` for ISO 8601 `Date` conversion into UTC +0 timezone, millisecond precision format, allowing it to accept a `Date` arg; this makes is much easier to convert `Date` values into the proper format with the correct timezone set and also preserves the existing functionality of using the current `Date()` as the default value when no arg is passed.

## Motivation and Context

3rd party extensions need to be able to convert `Date` instances into the ISO 8601 UTC +0 timezone, millisecond precision, and the current method only allows for converting the current `Date()` into this format. The alternative general purpose method can be cumbersome to use, and it would be easier to have the guarantee of setting the correct date format string and timezone using the existing method.

## How Has This Been Tested?

There is a new unit test case that validates the conversion of a Date instance into the ISO 8601 UTC +0 timezone, millisecond precision format using this method.

## Types of changes

## Checklist:

### DIFF
--- a/code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/TimeUtils.kt
+++ b/code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/TimeUtils.kt
@@ -47,15 +47,21 @@ object TimeUtils {
     }
 
     /**
-     * Gets the the ISO 8601 formatted date `String` for the current date.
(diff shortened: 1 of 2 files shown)


kevinlind (commented):
Can you apply the same change to `fun getIso8601Date()` so that a Date may be passed to it as well?

praveek:
Closing this in favour of #288

PR #288, opened 2022-12-02
TimeUtils Refactor

timkimadobe:
## Description

This PR refactors the existing `TimeUtils` to use a public `enum` system with a set of available date formatting patterns (the current set is pulled from the existing private formatting strings).

Both ISO 8601 helper methods have updated names to better reflect their return values, and both are able to accept an optional `Date` arg to convert using the specified pattern.

`getIso8601Date` -> `getIso8601DateTimeZoneRFC822` 
* There is value in being more specific with the time zone formatting this method uses, as it will reduce potential ambiguity, and this naming convention aligns with the next method:

`getIso8601DateTimeZoneISO8601` -> `getIso8601DateTimeZoneUTC` 
* The "TimeZoneISO8601" is not as clear as "TimeZoneUTC" because ISO 8601 is the general date pattern standard, with many variations possible; UTC makes it more clear that all times will be converted to UTC +0 (with further explanation in the method's docs)

The general purpose date converter `getIso8601Date(date:, format:, timeZone:)` has been updated to accept only non-optional enum values as the date formatting pattern and also renamed:
`getIso8601Date` -> `getFormattedDate` 
* Rationale: since this method handles all date formats, even those not conforming to ISO 8601, its name should reflect that capability; also, even before the enum system was implemented, there was no restriction on the date pattern that could be passed as the arg

The naming convention (generally) for the enum values themselves is as follows:
\<date pattern standard\>\_**TIMEZONE**\_\<time zone format\>\_**PRECISION**\_\<smallest precision unit\>
ISO8601\_**TIMEZONE**\_RFC822\_**PRECISION**\_SECOND
ISO8601\_**TIMEZONE**\_ISO8601\_UTCZ\_**PRECISION**\_MILLISECOND

TimeZone "GMT" has been updated to "Etc/UTC" to move off the [deprecated 3 letter time zones](https://developer.android.com/reference/java/util/TimeZone#three-letter-time-zone-ids), and "Etc/UTC" comes from the canonical value for UTC +0 (zero offset) from the [tzdb](https://en.wikipedia.org/wiki/Tz_database), supported by ICANN.

## Motivation and Context

This change would make it much easier for 3rd party extensions to get the common date formatting capabilities they need, through both: 
1. Clearly named helper/convenience methods that can convert passed `Date`s or the current `Date()` into commonly needed formats (automatically applying common built-in assumptions about usage; for example, the time zone to use)
    * `getIso8601DateTimeZoneRFC822` -> 2022-11-30T11:53:09-0700
    * `getIso8601DateTimeZoneUTC` -> 2022-11-30T18:53:09.497Z
2. A general purpose date converter that only accepts date patterns defined in the `enum` `TimeUtils.DatePattern`.
    * This allows for easy use of common date formatting patterns that have **guaranteed consistency** (as tested by the TimeUtils test cases, etc.), virtually eliminating potentials for typos, incorrect date formats, etc.

Point 2 also removes the burden of handling every potential date format pattern (that is, basically being a restrictively simple wrapper around `SimpleDateFormat`); it handles all the `enum` cases covered in `DatePattern` gracefully (for example, setting the time zone to the appropriate value automatically for the `ISO8601_TIMEZONE_ISO8601_UTCZ_PRECISION_MILLISECOND` pattern), and if users want their own custom format they can simply use the `SimpleDateFormat` directly.

Also these changes would be most optimal to decide on before the TimeUtils APIs are cemented in a public release.

## How Has This Been Tested?

Unit tests for the TimeUtils have been updated to reflect the new `enum` system, with a new test case for conversion of provided dates into the helper/convenience methods.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/TimeUtils.kt
+++ b/code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/TimeUtils.kt
@@ -21,7 +21,9 @@ import java.util.TimeZone
 object TimeUtils {
     private const val MILLISECONDS_PER_SECOND = 1000L
     private const val ISO8601_TIMEZONE_ISO8601_UTCZ_PRECISION_MILLISECOND = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
-    private const val ISO8601_TIMEZONE_RFC822_PRECISION_SECOND = "yyyy-MM-dd'T'HH:mm:ssZ"
+    private const val ISO8601_TIMEZONE_ISO8601_3X_PRECISION_SECOND = "yyyy-MM-dd'T'HH:mm:ssXXX"
+    private const val ISO8601_TIMEZONE_ISO8601_2X_PRECISION_SECOND = "yyyy-MM-dd'T'HH:mm:ssXX"
+    private const val ISO8601_FULL_DATE = "yyyy-MM-dd"
     private const val RFC2822_DATE_PATTERN = "EEE, dd MMM yyyy HH:mm:ss z"
 
     /**
@@ -35,24 +37,43 @@ object TimeUtils {
     }
 
     /**
-     * Gets the ISO 8601 formatted with RFC 822 time zone, second precision date `String` for the provided
-     * date using the current device time zone.
-     * Date pattern used is [ISO8601_TIMEZONE_RFC822_PRECISION_SECOND]
-     * which has timezone RFC 822 'Z' pattern, which gives a timezone of ex: "-0700" ("+0000" for UTC +0)
-     * Ex (device in time zone "America/Los_Angeles"): Wed Nov 30 11:53:09.497 GMT-07:00 2022 -> 2022-11-30T11:53:09-0700
+     * Gets the ISO 8601 formatted with colon time zone, second precision date `String`
+     * for the provide date using the system local time zone.
+     *
+     * Date pattern used is [ISO8601_TIMEZONE_ISO8601_3X_PRECISION_SECOND]
+     * which has timezone ISO 8601 'XXX' pattern, which gives a timezone of ex: "-07:00" ("Z" for UTC +0)
+     * Ex: (device in time zone "America/Los_Angeles") Wed Nov 30 11:53:09.497 GMT-07:00 2022 -> 2022-11-30T11:53:09-07:00
      *
      * @param date the [Date] to apply the formatting to; defaults to the current [Date]
-     * @return date [String] formatted as ISO 8601 with RFC 822 time zone, second precision
+     * @return date [String] formatted as ISO 8601 with colon time zone, second precision
      */
     @JvmStatic
     @JvmOverloads
-    fun getIso8601DateTimeZoneRFC822(date: Date = Date()): String {
-        return getFormattedDate(date, ISO8601_TIMEZONE_RFC822_PRECISION_SECOND) ?: ""
+    fun getISO8601Date(date: Date = Date()): String {
+        return getFormattedDate(date, ISO8601_TIMEZONE_ISO8601_3X_PRECISION_SECOND) ?: ""
+    }
+
+    /**
+     * Gets the ISO 8601 formatted with no colon time zone, second precision date `String`
+     * for the provide date using the system local time zone.
+     *
+     * Date pattern used is [ISO8601_TIMEZONE_ISO8601_2X_PRECISION_SECOND]
+     * which has timezone ISO 8601 'XX' pattern, which gives a timezone of ex: "-0700" ("Z" for UTC +0)
+     * Ex: (device in time zone "America/Los_Angeles") Wed Nov 30 11:53:09.497 GMT-07:00 2022 -> 2022-11-30T11:53:09-0700
+     *
+     * @param date the [Date] to apply the formatting to; defaults to the current [Date]
+     * @return date [String] formatted as ISO 8601 with no colon time zone, second precision
+     */
+    @JvmStatic
+    @JvmOverloads
+    fun getISO8601DateNoColon(date: Date = Date()): String {
+        return getFormattedDate(date, ISO8601_TIMEZONE_ISO8601_2X_PRECISION_SECOND) ?: ""
     }
 
     /**
      * Gets the ISO 8601 formatted with UTC(Z) time zone, millisecond precision date `String` for the
      * provided date using the UTC +0 time zone.
+     *
      * Date pattern used is [ISO8601_TIMEZONE_ISO8601_UTCZ_PRECISION_MILLISECOND]
      * which has timezone ISO 8601 'Z' char terminator, which gives a timezone of 'Z'.
      * Ex: Wed Nov 30 11:53:09.497 GMT-07:00 2022 -> 2022-11-30T18:53:09.497Z (notice the hour shift because the date must be evaluated from UTC +0)
@@ -71,14 +92,29 @@ object TimeUtils {
     }
 
     /**
-     * Gets the formatted date `String` for the passed in date; if no date passed in, uses current `Date()`
+     * Gets the ISO 8601 formatted full date `String` for the provided date using the system local time zone.
+     *
+     * Date pattern used is [ISO8601_FULL_DATE]
+     * Ex: Wed Nov 30 11:53:09.497 GMT-07:00 2022 -> 2022-11-30
      *
      * @param date the [Date] to apply the formatting to; defaults to the current [Date]
+     * @return date [String] formatted as ISO 8601 full date, using system local time zone
+     */
+    @JvmStatic
+    @JvmOverloads
+    fun getISO8601FullDate(date: Date = Date()): String {
+        return getFormattedDate(date, ISO8601_FULL_DATE) ?: ""
+    }
+
+    /**
+     * Gets the formatted date `String` for the passed in date; if no date passed in, uses current `Date()`
+     *
+     * @param date the [Date] to apply the formatting to
      * @param pattern the pattern [String] to use to format the date
      * @param timeZone the [TimeZone] to evaluate the formatted date from; defaults to device time zone if not specified
      * @return the formatted date [String], null if formatting fails
      */
-    private fun getFormattedDate(date: Date?, pattern: String, timeZone: TimeZone? = null): String? {
+    private fun getFormattedDate(date: Date, pattern: String, timeZone: TimeZone? = null): String? {
         // AMSDK-8374 -
         // we should explicitly ignore the device's locale when formatting an ISO 8601 timestamp
         val posixLocale = Locale(Locale.US.language, Locale.US.country, "POSIX")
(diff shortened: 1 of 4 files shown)


code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/TimeUtils.kt
```diff
@@ -20,10 +20,13 @@ import java.util.TimeZone
 
 object TimeUtils {
     private const val MILLISECONDS_PER_SECOND = 1000L
-    private const val ISO8601_DATE_FORMATTER_TIMEZONE_RFC822 = "yyyy-MM-dd'T'HH:mm:ssZZZ"
-    private const val ISO8601_DATE_FORMATTER_TIMEZONE_ISO8601 = "yyyy-MM-dd'T'HH:mm:ssXXX"
-    private const val RFC2822_DATE_PATTERN = "EEE, dd MMM yyyy HH:mm:ss z"
-    private const val ISO8601_DATE_FORMATTER_TIMEZONE_UTC = "yyyy-MM-dd'T'HH:mm:ss.SSS'Z'"
+
+    enum class DatePattern(val pattern: String) {
```
praveek, line 24: I would prefer private string constants instead of enum. This will allow other extensions to use `getFormattedString` even with formats not declared here.

code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/TimeUtils.kt
```diff
@@ -36,45 +39,69 @@ object TimeUtils {
...
      *
-     * @return Iso8601 formatted date [String]
+     * @param date the [Date] to apply the formatting to; defaults to the current [Date]
+     * @return date [String] formatted as ISO 8601 with RFC 822 time zone, second precision
      */
     @JvmStatic
-    fun getIso8601Date(): String? {
-        return getIso8601Date(Date(), ISO8601_DATE_FORMATTER_TIMEZONE_RFC822)
+    @JvmOverloads
+    fun getIso8601DateTimeZoneRFC822(date: Date? = Date()): String? {
```
praveek, line 53: Can you make data non optional?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/TimeUtils.kt
```diff
@@ -36,45 +39,69 @@ object TimeUtils {
...
-     * @return Iso8601 formatted date [String]
+     * @param date the [Date] to apply the formatting to; defaults to the current [Date]
+     * @return date [String] formatted as ISO 8601 with UTC(Z) time zone, millisecond precision
      */
     @JvmStatic
-    fun getIso8601DateTimeZoneISO8601(): String? {
-        return getIso8601Date(Date(), ISO8601_DATE_FORMATTER_TIMEZONE_UTC, TimeZone.getTimeZone("GMT"))
+    @JvmOverloads
+    fun getIso8601DateTimeZoneUTC(date: Date? = Date()): String? {
+        return getFormattedDate(date, DatePattern.ISO8601_TIMEZONE_ISO8601_UTCZ_PRECISION_MILLISECOND, TimeZone.getTimeZone("Etc/UTC"))
```
praveek, line 75: Any specific reason to change timezone?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/TimeUtils.kt
```diff
@@ -36,45 +39,69 @@ object TimeUtils {
...
-     * @return ISO 8601 formatted date [String]
+     * @param date the [Date] to apply the formatting to; defaults to the current [Date]
+     * @param datePattern the [DatePattern] to use to format the date [String]
+     * @param timeZone the [TimeZone] to evaluate the formatted date from; defaults to device time zone if not specified
+     * @return the formatted date [String]
      */
     @JvmStatic
     @JvmOverloads
-    fun getIso8601Date(date: Date?, format: String?, timeZone: TimeZone? = null): String? {
+    fun getFormattedDate(date: Date?, datePattern: DatePattern, timeZone: TimeZone? = null): String? {
```
praveek, line 88: Can you make this private for now? We can expose this later if any extension needs access to it.

praveek: Remove these annotations for non public methods.

code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/TimeUtils.kt
```diff
@@ -36,45 +39,69 @@ object TimeUtils {
...
      *
-     * @return Iso8601 formatted date [String]
+     * @param date the [Date] to apply the formatting to; defaults to the current [Date]
+     * @return date [String] formatted as ISO 8601 with UTC(Z) time zone, millisecond precision
      */
     @JvmStatic
-    fun getIso8601DateTimeZoneISO8601(): String? {
-        return getIso8601Date(Date(), ISO8601_DATE_FORMATTER_TIMEZONE_UTC, TimeZone.getTimeZone("GMT"))
+    @JvmOverloads
+    fun getIso8601DateTimeZoneUTC(date: Date? = Date()): String? {
```
praveek, line 74: Can you add WithMillisecond suffix to this method?

praveek:
Can you also fix lint errors and update the apidump file.

code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/TimeUtils.kt
emdobrin: I also suggest we name these methods similar to the swift ones e.g. `getISO8601UTCDateWithMilliseconds` - ref https://github.com/adobe/aepsdk-core-ios/blob/main/AEPServices/Sources/utility/Date%2BFormat.swift

code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/TimeUtils.kt
praveek: Can you change the implementation to 
```
fun getIso8601DateTimeZoneRFC822(date: Date = Date()): String {
	return getFormattedDate(date,DatePattern.ISO8601_TIMEZONE_RFC822_PRECISION_SECOND) ?? ""
}
```
Also make similar changes to other functions

code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/TimeUtils.kt
```diff
@@ -36,45 +39,66 @@ object TimeUtils {
...
      *
-     * @return Iso8601 formatted date [String]
+     * @param date the [Date] to apply the formatting to; defaults to the current [Date]
+     * @return date [String] formatted as ISO 8601 with UTC(Z) time zone, millisecond precision
      */
     @JvmStatic
-    fun getIso8601DateTimeZoneISO8601(): String? {
-        return getIso8601Date(Date(), ISO8601_DATE_FORMATTER_TIMEZONE_UTC, TimeZone.getTimeZone("GMT"))
+    @JvmOverloads
+    fun getISO8601UTCDateWithMilliseconds(date: Date? = Date()): String? {
```
praveek, line 73: Similar change as above

code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/TimeUtils.kt
```diff
@@ -36,45 +39,66 @@ object TimeUtils {
...
-     * @return ISO 8601 formatted date [String]
+     * @param date the [Date] to apply the formatting to; defaults to the current [Date]
+     * @param datePattern the [DatePattern] to use to format the date [String]
+     * @param timeZone the [TimeZone] to evaluate the formatted date from; defaults to device time zone if not specified
+     * @return the formatted date [String]
      */
-    @JvmStatic
-    @JvmOverloads
-    fun getIso8601Date(date: Date?, format: String?, timeZone: TimeZone? = null): String? {
+    private fun getFormattedDate(date: Date?, datePattern: DatePattern, timeZone: TimeZone? = null): String? {
```
praveek, line 85:
```
private fun getFormattedDate(date: Date, pattern: String, timeZone: TimeZone? = null): String? ```

File: code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/TimeUtils.kt [hunk 8]
```diff
@@ -36,45 +39,66 @@ object TimeUtils {
...
         // AMSDK-8374 -
         // we should explicitly ignore the device's locale when formatting an ISO 8601 timestamp
         val posixLocale = Locale(Locale.US.language, Locale.US.country, "POSIX")
-        val iso8601Format: DateFormat = SimpleDateFormat(format, posixLocale)
-        if (timeZone != null) {
-            iso8601Format.timeZone = timeZone
+        val dateFormat: DateFormat = SimpleDateFormat(datePattern.pattern, posixLocale)
+
+        // Handle special cases/requirements based on DatePattern provided
+        when (datePattern) {
```
praveek, line 92: Let us remove this additional logic and make it the responsibility of caller to set the correct timezone.

praveek:
Please fix the build error in tests

praveek (approved):
LGTM. Appreciate the detailed function descriptions.

code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/TimeUtils.kt
```diff
@@ -98,7 +147,7 @@ object TimeUtils {
 
     /**
      * Converts the epoch into a RFC-2822 date string pattern - [RFC2822_DATE_PATTERN]
-     * @param epoch the epoch that should be converted to Date
+     * @param epoch the epoch (milliseconds) that should be converted to [Date]
      * @param timeZone the timezone that should be used
      * @param locale the locale whose date format symbols should be used
      * @return a RFC-2822 formatted date string for the epoch provided
```
emdobrin, line 153: @praveek on getRFC2822Date - I haven't seen this one in iOS, are there any extensions using it or are we using it for logging, or can it be removed?

emdobrin (commented):
Latest changes look great. I left a small question

code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/TimeUtils.kt
prudrabhat: This is used for the purposes of storing date from responses in cache metadata. Core 2.0 cache system uses a file backed cache metadata. So in order to verify the last modified time of the cached file from the the one in the response we store the last modified in the cache metadata. I read that the epoch is the preferred way to store date/time over a specific formatted date in file backed cache metadata and hence this logic was added.

praveek:
@timkimadobe Can you resolve the conflicts so that we can merge this in?

Issue #92, opened 2022-06-09
Validate Lifecycle sets valid environment language string when building XDM metrics


kevinlind:
### Expected Behaviour
The Android 1.0 Lifecycle V2 Metrics Builder class does set a valid IETF BCP 47 locale string when the device's locale contains a script or extension tag. This causes issues because the XDM definition for environment._dc.language validates the string and rejects the data if it does not conform. The culprit is the use of Java's Locale.toString() API which doesn't return a valid IETF BCP 47 string but instead a simple string representation of the Locale object. 

This should get corrected in Android 2.0 when porting the Lifecycle logic.

The solution here is to not use `toString` for any case other than logging, and instead either 1) build a valid IETF BCP 47 string from the Locale object's language, region, and variant codes, or 2) if Java 7 is used (and minimum Android API 21) then `Locale.toLanguageTag()` can be used.

See internal Jira ticket PLATIR-20377 for more details.

References:
https://docs.oracle.com/javase/8/docs/api/
https://tools.ietf.org/search/bcp47
https://github.com/adobe/xdm/blob/master/components/datatypes/environment.schema.json#L211

PR #293, opened 2022-12-08
[MOB-17661] Lifecycle v2 set BCP 47 style language tag for XDM field environment._dc.language

kevinlind:
## Description

Format the value set in the XDM field environment._dc.language to a BCP 47 style language tag.
The XDM schema for Environment requires the _dc.language field be formatted to BCP 47. The existing format used in the v2 metrics builder did not use a valid format when the locale contained script, variant, or extension codes. 
Fixes the issue by:

- When using Android API 21+, use Locale.toLanguageTag() which returns a valid BCP 47 style tag
- When using lower Android APIs, formats the language tag by using _languageCode_-_countryCode_
- Adds a validation step in XDMLifecycleEnvironment as a failsafe so only valid language tags are added to environment._dc.language.

This PR does not change the locale format in any other uses than defined above.

## Related Issue
https://github.com/adobe/aepsdk-core-android/issues/92

## Types of changes

## Checklist:

### DIFF
--- a/code/android-lifecycle-library/src/main/java/com/adobe/marketing/mobile/lifecycle/LifecycleUtil.java
+++ b/code/android-lifecycle-library/src/main/java/com/adobe/marketing/mobile/lifecycle/LifecycleUtil.java
@@ -13,9 +13,7 @@ package com.adobe.marketing.mobile.lifecycle;
 
 import android.annotation.SuppressLint;
 import android.os.Build;
-
 import androidx.annotation.VisibleForTesting;
-
 import com.adobe.marketing.mobile.util.StringUtils;
 import java.text.SimpleDateFormat;
 import java.util.Date;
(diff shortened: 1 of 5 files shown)


code/android-lifecycle-library/src/main/java/com/adobe/marketing/mobile/lifecycle/LifecycleUtil.java
```diff
@@ -57,13 +62,56 @@ private static String dateToISO8601String(final Date timestamp, final String tim
...
+	 * Format the locale to the string format used in XDM.
+	 * For Android API version >= Lollipop (21), returns {@link Locale#toLanguageTag()}.
+	 * For Android API version < 21, returns a concatenation of {@link Locale#getLanguage()}
+	 * and {@link Locale#getCountry()}, separated by '-'.
+	 *
+	 * @param locale active Locale value
+	 * @return String representation of the locale
+	 */
+	@SuppressLint("NewApi")
+	static String formatLocaleXDM(final Locale locale) {
```
emdobrin, line 90: Changes look good overall, but I would not limit this API to the xdm field only but standardizing it in the SDK usages. Can we make this a new deviceInfo API `getActiveLocaleBCPString()` and remove any custom formatters in extensions code?

kevinlind: I like this idea as it provides consistent locale usage across the entire SDK. @praveek what do you think?

praveek: Yes. It sounds good as other extensions can reuse the implementation from DeviceInforming.

code/android-lifecycle-library/src/main/java/com/adobe/marketing/mobile/lifecycle/LifecycleUtil.java
```diff
@@ -59,12 +62,56 @@ private static String dateToISO8601String(final Date timestamp, final String tim
...
+     */
+    @SuppressLint("NewApi")
+    static String formatLocaleXDM(final Locale locale) {
+        if (locale == null) {
+            return null;
+        }
+
+        if (isLollipopOrGreater.check()) {
+            return locale.toLanguageTag();
+        } else {
```
prudrabhat, line 96: nit: logic in the `else` block can be pulled out (to avoid nesting if's )

prudrabhat (approved):

prudrabhat (approved):

code/android-lifecycle-library/src/main/java/com/adobe/marketing/mobile/lifecycle/LifecycleUtil.java
kevinlind: The current thinking is to change the logic in DeviceInforming.getLocaleString() to return a BCP-47 style string instead of adding a second API. We will need to vet this change with some stakeholders which consume the user-agent and accept-language headers as those formates will changes slightly. I'll take up this change in another PR.

PR #296, opened 2022-12-14
Quick fix: the first Identity shared state should always contain ecid

yangyansong-adbe:


prudrabhat (commented):
I don't know about the Identity workflow much so I will look to Praveen to review the flow/edge case. Can you add a small test to verify the change as well.

praveek (approved):

PR #297, opened 2022-12-15
Fix for issue (295) - methods needed for Messaging extension

yangyansong-adbe:


code/android-core-library/src/main/java/com/adobe/marketing/mobile/services/ui/FullscreenMessage.java
```diff
@@ -46,4 +48,18 @@ public interface FullscreenMessage {
      *     message.
      */
     void setMessageSetting(final MessageSettings messageSettings);
+
+    /**
+     * Return an instance of {@link WebView} setup for the current {@link FullscreenMessage}
+     *
+     * @return an instance of {@link WebView}
+     */
+    @Nullable WebView getWebView();
```
praveek, line 57: Can you remove @Nullable annotation as we don't set it for any other method?

code/android-core-library/src/main/java/com/adobe/marketing/mobile/util/UrlUtils.java
```diff
@@ -49,4 +51,39 @@ public static boolean isValidUrl(final String stringUrl) {
...
         return UrlEncoder.urlEncode(unencodedString);
     }
+
+    /**
+     * Extra URL parameters and return as a {@code Map}
+     *
+     * @param queryString the {@code String} of the URL parameter section
+     * @return a {@code Map} of URL parameters
+     */
+    public static Map<String, String> extractQueryParameters(final String queryString) {
```
praveek, line 61: Can you add some tests?

prudrabhat, line 61: What do you think of using the [android.net.Uri#getQueryParameterNames](https://developer.android.com/reference/android/net/Uri#getQueryParameters(java.lang.String)) and then use the [uri.getQueryParameter(<paramater_name>)](https://developer.android.com/reference/android/net/Uri#getQueryParameter(java.lang.String)) to build the result?  (we may need to have the callers pass the entire url/uri to use it but I think that is just fine) 
That way our code does not have any internal parsing logic on the url/uri.

yangyansong-adbe: It's ported from the 1.x utility class, and I didn't see any tests for it. Can we add tests in another PR to unlock the messaging migration work?

yangyansong-adbe: It looks like we don't need to expose this utility method in Core. Other extensions can use the above Android APIs to get the map they need.  @praveek @sbenedicadb thoughts?

prudrabhat: I think that there is still some value in exposing this utility as it is being using by 2 extensions. My intention was just to ensure the logic does not have additional parsing/splitting..

yangyansong-adbe: I think having several lines of duplicated logic in extensions should be fine. It should be easier to make changes for the public APIs in Core.

sbenedicadb (approved):
looks good pending the changes requested by other reviewers

praveek (approved):
Merge after addressing the comments.

prudrabhat (approved):

yangyansong-adbe:
Will add tests for `UrlUtils. extractQueryParameters` in another PR, here is the issue to track it -> https://github.com/adobe/aepsdk-core-android/issues/299

PR #313, opened 2023-01-04
[Mob 17736] Handle getExperienceCloudId and getIdentifiers events without valid configuration state

yangyansong-adbe:


code/integration-tests/src/androidTest/java/com/adobe/marketing/mobile/integration/identity/IdentityIntegrationTests.kt
```diff
@@ -463,6 +463,7 @@ class IdentityIntegrationTests {
         }
         MobileCore.resetIdentities()
         countDownLatchSecondNetworkMonitor.await()
+        Thread.sleep(20)
```
praveek, line 466: What is the need for such a long sleep here? Can you reduce the timeout?

praveek:
Can you add some test scenarios around this logic?

code/integration-tests/src/androidTest/java/com/adobe/marketing/mobile/integration/identity/IdentityIntegrationTests.kt
yangyansong-adbe: `mid` will be refreshed in the local storage after the network call, So just wait a bit here to get the refreshed `mid` value.

praveek (approved):

PR #326, opened 2023-01-12
Add docs for "Getting started" and "Migration" sections

yangyansong-adbe:


docs/Migration.md
```diff
@@ -0,0 +1,33 @@
...
+## Upgrading
+
+### Gradle
+
+To upgrade, update the Gradle dependency references to `Android SDK 2.0` in the Android app's `build.gradle`:
+
+```diff
+- implementation 'com.adobe.marketing.mobile:sdk-core:1.+'
+
++ implementation 'com.adobe.marketing.mobile:core:2.0.0'
```
praveek, line 14: Can you also add a note telling them to update gradle dependencies to latest available versions

docs/GettingStarted.md
```diff
@@ -0,0 +1,135 @@
...
+        MobileCore.setApplication(this);
+        MobileCore.setLogLevel(LoggingMode.VERBOSE);
+        MobileCore.configureWithAppID(APP_ID);
+
+        List<Class<? extends Extension>> extensions = new ArrayList<>();
+        extensions.add(Lifecycle.EXTENSION);
+        extensions.add(Signal.EXTENSION);
+        extensions.add(UserProfile.EXTENSION);
+        extensions.add(Identity.EXTENSION);
+        MobileCore.registerExtensions(extensions, o -> {
```
prudrabhat, line 48: Can you change this to the following so that it is consistent with the Kotlin example ?
```suggestion 
final List<Class<? extends Extension>> extensions = Arrays.asList(Lifecycle.EXTENSION, Signal.EXTENSION, ...);
```

docs/GettingStarted.md
```diff
@@ -0,0 +1,135 @@
...
+## Get the Experience Platform SDK
+
+If you cannot access the Mobile Install Instructions dialog box in the Data Collection UI, complete the following sections to get the Adobe Experience Platform SDK. If you already completed the steps in the Mobile Install Instructions dialog box, no need to complete these steps.
+
+## Installation instructions
+
+1. Each extension needs to be added as a dependency to the mobile application project. The following examples will add the Mobile Core, Lifecycle, Identity, Signal and Profile extensions.
+
+```java
+implementation 'com.adobe.marketing.mobile:userprofile:2.0'
```
praveek, line 20: Can you also replace the version numbers in this section to 2.x.x and add the same note as below 
> The following 2.x.x should be replaced with the latest version numbers.

docs/GettingStarted.md
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+}
+```
+
+## Implement Lifecycle Metrics
+
+Lifecycle metrics is an optional, yet valuable feature provided by the Adobe Experience Platform SDK. It offers out-of-the-box, application lifecycle information about an app user. These metrics contain information on the app user's engagement lifecycle, such as device information, install or upgrade information, session start and pause times, and more.
+
+Withint the `Activity.onResume()` function, start Lifecycle data collection:
```
prudrabhat, line 79: _within_ the ..

docs/GettingStarted.md
```diff
@@ -0,0 +1,135 @@
...
+```kotlin
+override fun onResume() {
+    super.onResume()
+    MobileCore.setApplication(application)
+    MobileCore.lifecycleStart(null)
+}
+```
+
+Setting the application is only necessary on activities that are entry points for your application. However, setting the application on each Activity has no negative impact and ensures that the SDK always has the necessary reference to your application. As a result, you should call setApplication on each of your activities.
+You can use the onPause function to pause the lifecycle data collection:
```
prudrabhat, line 103: Can you make the api names to be in code format.
example : You can use the `onPause` ... , As a result, you should call `setApplication` ...

docs/Migration.md
```diff
@@ -0,0 +1,35 @@
+# Migrate an existing Android app to AEP Android SDK 2.0
+
+This document describes how existing Android SDK 1.x customers can upgrade to the Android SDK 2.0 without introducing any breaking changes. We have made upgrading seamless, which will ensure that APIs between SDK 1.x and 2.0 are compatible.
```
prudrabhat, line 3: nit: I wonder if calling the sdk Android SDK 2.0 will cause confusion for implementers. I cannot think of another name that fits here though! 😅 . Do you think adding a maven hyperlink to these SDK's will clarify it? Your call.

docs/GettingStarted.md
```diff
@@ -0,0 +1,135 @@
...
+1. Each extension needs to be added as a dependency to the mobile application project. The following examples will add the Mobile Core, Lifecycle, Identity, Signal and Profile extensions.
+
+```java
+implementation 'com.adobe.marketing.mobile:userprofile:2.0'
+implementation 'com.adobe.marketing.mobile:core:2.0'
+implementation 'com.adobe.marketing.mobile:identity:2.0'
+implementation 'com.adobe.marketing.mobile:signal:2.0'
+implementation 'com.adobe.marketing.mobile:lifecycle:2.0'
+```
+
```
prudrabhat, line 26: Should we add a note asking implementers to check the latest version of dependancies ?

docs/GettingStarted.md
```diff
@@ -0,0 +1,135 @@
...
+2. Next you'll need to import SDK libraries into your project and register them for initialization. Extensions are registered with Mobile Core so that they can dispatch and listen for events.
+
+### Java
+
+```java
+public class MainApp extends Application {
+    private static final String APP_ID = "";
```
prudrabhat, line 33:
```suggestion
    private static final String APP_ID = "YOUR_APP_ID";
```

docs/Migration.md
yangyansong-adbe: We will show maven links for core extensions on the main README. So, I think it should be good.

praveek (approved):

PR #355, opened 2023-01-27
Ensure initial shared state is created once and sync new ECID immediately after reset

kevinlind:
## Description

1. There are cases when Identity will create an initial shared state using “null” twice when booting. The iOS implementation has a check for this which I’ll need to add to the Android implementation.
2. On privacy toggle or a call to resetIdentities, a force sync event is dispatched to the Event Hub which when processed will generate a new ECID. This can cause a delay where, if any other events are already in the queue will get processed without an ECID. iOS calls syncIdentifiers directly here to ensure the ECID is generated when resetIdentities is processed. I’ll make that change to Android as well.

## Types of changes

## Checklist:

### DIFF
--- a/code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
+++ b/code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
@@ -154,7 +154,7 @@ public final class IdentityExtension extends Extension {
         getApi().registerEventListener(
                         EventType.GENERIC_IDENTITY,
                         EventSource.REQUEST_RESET,
-                        this::processIdentityRequest);
+                        this::handleIdentityRequestReset);
         // listen to Analytics response
         getApi().registerEventListener(
                         EventType.ANALYTICS,
@@ -467,16 +467,7 @@ public final class IdentityExtension extends Extension {
      *
      * @param event the request request {@link Event}
      */
-    void handleIdentityRequestReset(final Event event) {
-        if (event == null) {
-            Log.debug(
-                    IdentityConstants.LOG_TAG,
-                    LOG_SOURCE,
-                    LOG_SOURCE,
-                    "handleIdentityRequestReset: Ignoring null event");
-            return;
-        }
-
+    void handleIdentityRequestReset(@NonNull final Event event) {
         if (privacyStatus == MobilePrivacyStatus.OPT_OUT) {
             Log.debug(
                     IdentityConstants.LOG_TAG,
@@ -763,9 +754,7 @@ public final class IdentityExtension extends Extension {
                 "processEvent : Processing the Identity event: %s",
                 event);
 
-        if (isResetIdentityEvent(event)) {
-            handleIdentityRequestReset(event);
-        } else if (isSyncEvent(event) || event.getType().equals(EventType.GENERIC_IDENTITY)) {
+         if (isSyncEvent(event) || event.getType().equals(EventType.GENERIC_IDENTITY)) {
             if (handleSyncIdentifiers(event, configSharedState, false)) {
                 getApi().createSharedState(packageEventData(), event);
             }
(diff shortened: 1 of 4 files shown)


yangyansong-adbe (approved):

code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
```diff
@@ -478,14 +497,21 @@ void handleIdentityRequestReset(final Event event) {
         savePersistently(); // clear datastore
```
emdobrin, line 494: can simplify the code if using #NotNull event for handlers

code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
```diff
@@ -1884,11 +1898,23 @@ void processPrivacyChange(final Event event, final Map<String, Object> eventData
...
+                    getApi().getSharedState(
+                                    IdentityConstants.EventDataKeys.Configuration.MODULE_NAME,
+                                    event,
+                                    false,
+                                    SharedStateResolution.LAST_SET);
+            if (configState != null) {
+                configSharedState.getConfigurationProperties(configState.getValue());
+            }
+
+            if (handleSyncIdentifiers(event, configSharedState, true)) {
```
emdobrin, line 1904: for both processPrivacyChange and resetIdentities ios calls syncIdentifiers(event:event) -> with default forceSync false 

https://github.com/adobe/aepsdk-core-ios/blob/main/AEPIdentity/Sources/IdentityState.swift#L348

code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
```diff
@@ -822,7 +841,7 @@ boolean handleSyncIdentifiers(
                     "handleSyncIdentifiers : Ignoring the Sync Identifiers call because the"
                             + " privacy status was opt-out.");
             // did process this event but can't sync the call. Hence return true.
```
emdobrin, line 843: nit: comment is outdated

code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
```diff
@@ -864,7 +883,7 @@ boolean handleSyncIdentifiers(
                     LOG_SOURCE,
                     "handleSyncIdentifiers : Ignored the Sync Identifiers call because the privacy"
                             + " status was opt-out.");
-            return true; // cannot process event due to privacy setting, remove from queue
+            return false;
```
emdobrin, line 886: not sure why we handle privacy again here. we should update L837 to get privacy from latest config ss and remove this repeating check L879->L886

code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
kevinlind: If the ECID is null, then that will force sync, so removing the "forceSync" flag is okay.

code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
kevinlind: It's the same as in Android 1.x.  iOS cleaned this up a bit to only use "latestValidConfig". That will require a bit more changes to the code and test cases, so I'll defer that to another PR.

code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
```diff
@@ -763,9 +754,7 @@ void processIdentityRequest(@NonNull final Event event) {
                 "processEvent : Processing the Identity event: %s",
                 event);
 
-        if (isResetIdentityEvent(event)) {
-            handleIdentityRequestReset(event);
-        } else if (isSyncEvent(event) || event.getType().equals(EventType.GENERIC_IDENTITY)) {
+         if (isSyncEvent(event) || event.getType().equals(EventType.GENERIC_IDENTITY)) {
```
emdobrin: let's reverse this one, using the constant first

emdobrin (approved):

emdobrin:
:shipit:

PR #399, opened 2023-02-24
Fix the flaky tests

yangyansong-adbe:


code/core/src/test/java/com/adobe/marketing/mobile/MobileCoreTests.kt
```diff
@@ -19,17 +19,28 @@ import com.adobe.marketing.mobile.internal.eventhub.EventHubError
...
+import org.junit.runner.RunWith
+import org.mockito.Mock
 import org.mockito.Mockito.mock
+import org.mockito.Mockito.reset
+import org.mockito.Mockito.verifyNoInteractions
+import org.mockito.junit.MockitoJUnitRunner
+import org.mockito.kotlin.KArgumentCaptor
+import org.mockito.kotlin.argumentCaptor
+import org.mockito.kotlin.times
+import org.mockito.kotlin.verify
```
praveek, line 31: The changes look good but can you separate the registration/initialization tests and logic which uses real EventHub into MobileCoreRegistrationTests.kt. We can keep all the API tests in this file and move `EventHub.shared = mockedEventHub` into the setup() function.

praveek (approved):

PR #416, opened 2023-03-23
Use latest configuration on Identity bootup (resolves #414)

### LINKED ISSUE #414: Identity should use last set valid config on bootup
Consolidate the config retrieval for bootup scenario. Currently it is being retrieved twice, once for last set value and once with last set value for current event, which may return different results or block on incorrect configuration.
Ref: IdentityExtension forceSyncIdentifiers, readyForSyncIdentifiers.

kevinlind:
## Description

Currently, Identity uses the first set Configuration shared state when processing the force sync during boot up. This causes issues when the first Configuration is not valid. Instead, the Identity boot sequence should use the last set Configuration shared state when processing the force sync event. This way, if the first set Configuration state is invalid, then the next set Configuration state can be used to initialize the Identity extension.

- Refactor `readyForSyncIdentifiers` to take in a Map of the configuration state instead of getting the configuration state from an Event. In addition, update the `latestValidConfig` if the configuration state contains a `experienceCloud.org` value. This is similar to the iOS Swift implementation.
- Update `forceSyncIdentifiers` to use the refactored `readyForSyncIdentifiers`, ensuring to pass `null` to `getSharedState()` which will retrieve the last set Configuration state.
- Refactor `handleSyncIdentifiers` to remove the configuration parameter. As `latestValidConfig` is updated for each sync call, `handleSyncIdentifiers` just needs to use it for the configuration state. This is similar to the iOS Swift implementation.

## Types of changes

## Checklist:

### DIFF
--- a/code/identity/src/main/java/com/adobe/marketing/mobile/identity/ConfigurationSharedStateIdentity.java
+++ b/code/identity/src/main/java/com/adobe/marketing/mobile/identity/ConfigurationSharedStateIdentity.java
@@ -35,12 +35,12 @@ final class ConfigurationSharedStateIdentity {
     private final MobilePrivacyStatus privacyStatus;
 
     /** {@link String} containing the host value for the Experience Cloud Server */
-    private final String marketingCloudServer;
+    private final String experienceCloudServer;
 
     /**
      * Extracts data from the provided shared state data and sets the internal configuration
      * appropriately. Sets default values for {@link #privacyStatus}, and {@link
-     * #marketingCloudServer} if provided shared state does not contain valid values.
+     * #experienceCloudServer} if provided shared state does not contain valid values.
      *
      * @param sharedState EventData representing a {@code Configuration} shared state
      */
@@ -53,11 +53,7 @@ final class ConfigurationSharedStateIdentity {
                         IdentityConstants.JSON_EXPERIENCE_CLOUD_SERVER_KEY,
                         Defaults.SERVER);
 
-        if (StringUtils.isNullOrEmpty(server)) {
-            this.marketingCloudServer = Defaults.SERVER;
-        } else {
-            this.marketingCloudServer = server;
-        }
+        this.experienceCloudServer = StringUtils.isNullOrEmpty(server) ? Defaults.SERVER : server;
 
         this.privacyStatus =
                 MobilePrivacyStatus.fromString(
@@ -90,8 +86,8 @@ final class ConfigurationSharedStateIdentity {
      *
      * @return the Experience Cloud server; defaults to {@link Defaults#SERVER}
      */
-    @NonNull String getMarketingCloudServer() {
-        return marketingCloudServer;
+    @NonNull String getExperienceCloudServer() {
+        return experienceCloudServer;
     }
 
     /**
--- a/code/identity/src/main/java/com/adobe/marketing/mobile/identity/ConfigurationSharedStateIdentity.java
+++ b/code/identity/src/main/java/com/adobe/marketing/mobile/identity/ConfigurationSharedStateIdentity.java
@@ -11,9 +11,10 @@
 
 package com.adobe.marketing.mobile.identity;
 
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
 import com.adobe.marketing.mobile.MobilePrivacyStatus;
 import com.adobe.marketing.mobile.identity.IdentityConstants.Defaults;
-import com.adobe.marketing.mobile.services.Log;
 import com.adobe.marketing.mobile.util.DataReader;
 import com.adobe.marketing.mobile.util.StringUtils;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
```diff
@@ -860,31 +844,22 @@ boolean handleSyncIdentifiers(
...
+            // sanity check, should never get here as latestValidConfig set from
+            // readyForEvent/readyForSyncIdentifiers
+            // can't process this event. return false to break execution loop
+            Log.debug(
+                    IdentityConstants.LOG_TAG,
+                    LOG_SOURCE,
+                    "handleSyncIdentifiers : Unable to process sync identifiers request as the"
+                            + " configuration did not contain a valid Experience Cloud"
+                            + " organization ID. Will attempt to process event when a valid"
+                            + " configuration is received.");
```
emdobrin, line 857: Here there will be no other attempt for this request, let's adjust the log comment to capture that:
```
"handleSyncIdentifiers: Ignoring the Sync Identifiers call as last valid config is empty."
```

code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
```diff
@@ -860,31 +844,22 @@ boolean handleSyncIdentifiers(
...
-                        IdentityConstants.LOG_TAG,
-                        LOG_SOURCE,
-                        "handleSyncIdentifiers : Unable to process sync identifiers request as the"
-                                + " configuration did not contain a valid Experience Cloud"
-                                + " organization ID. Will attempt to process event when a valid"
-                                + " configuration is received.");
-                return false;
-            }
+        if (latestValidConfig == null) {
+            // sanity check, should never get here as latestValidConfig set from
```
emdobrin, line 848: nit: this should not have any impact on the flow, but for consistency with iOS can we do this sanity check first thing in this method?

code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
```diff
@@ -815,32 +813,18 @@ void handleOptOut(final Map<String, Object> configuration) {
...
-                    IdentityConstants.LOG_TAG,
-                    LOG_SOURCE,
-                    "handleSyncIdentifiers : Ignoring the Sync Identifiers call because the"
-                            + " configuration was null.");
-            return false;
-        }
-
+    boolean handleSyncIdentifiers(final Event event, final boolean forceSync) {
         // do not even extract any data if the config is opt_out.
         if (privacyStatus == MobilePrivacyStatus.OPT_OUT) {
```
emdobrin, line 829: shouldn't we use latestValidConfig.privacyStatus instead and remove L861->L869?

code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
```diff
@@ -894,14 +869,14 @@ boolean handleSyncIdentifiers(
         }
 
         // if the marketingCloudServer is null or empty use the default server
-        if (StringUtils.isNullOrEmpty(currentEventValidConfig.marketingCloudServer)) {
-            currentEventValidConfig.marketingCloudServer = IdentityConstants.Defaults.SERVER;
+        if (StringUtils.isNullOrEmpty(latestValidConfig.marketingCloudServer)) {
+            latestValidConfig.marketingCloudServer = IdentityConstants.Defaults.SERVER;
```
emdobrin, line 873: I find it odd that we update properties of latestValidConfig as part of this method. Instead ConfigurationSharedStateIdentity should take care of replacing null/empty values with the default server

code/integration-tests/src/androidTest/java/com/adobe/marketing/mobile/integration/identity/IdentityIntegrationTests.kt
```diff
@@ -216,6 +216,74 @@ class IdentityIntegrationTests {
...
     }
 
+    @Test(timeout = 5_000)
+    fun testIdentitySendsForceSyncRequestOnLaunchOnValidConfig() {
+        SDKHelper.resetSDK()
+        ServiceProviderModifier.reset()
+
+        overrideNetworkService()
+
+        MobileCore.setApplication(ApplicationProvider.getApplicationContext())
```
emdobrin, line 226: should we cover these in the setup instead?

code/integration-tests/src/androidTest/java/com/adobe/marketing/mobile/integration/identity/IdentityIntegrationTests.kt
```diff
@@ -216,6 +216,74 @@ class IdentityIntegrationTests {
...
+        MobileCore.setLogLevel(LoggingMode.VERBOSE)
```
emdobrin, line 227: used for debugging or can it be removed?

code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
kevinlind: ConfigurationSharedStateIdentity already sets marketingCloudServer to the default if null/empty, so this check is not needed. I went ahead and made this class immutable as it's odd to have a class like this which is mutable.

code/integration-tests/src/androidTest/java/com/adobe/marketing/mobile/integration/identity/IdentityIntegrationTests.kt
kevinlind: I refactored this test case. I originally thought I needed to restart the SDK, but I only needed to set the configuration and allow setUp to register the extensions for me.

code/identity/src/main/java/com/adobe/marketing/mobile/identity/ConfigurationSharedStateIdentity.java
```diff
@@ -83,6 +67,33 @@ void getConfigurationProperties(final Map<String, Object> sharedState) {
...
+    @NonNull MobilePrivacyStatus getPrivacyStatus() {
+        return privacyStatus;
+    }
+
+    /**
+     * The Experience Cloud server.
+     *
+     * @return the Experience Cloud server; defaults to {@link Defaults#SERVER}
+     */
+    @NonNull String getMarketingCloudServer() {
```
emdobrin, line 93: since you did a bunch of changes in this class, can we also rename these methods and properties marketingCloudServer -> experienceCloudServer

code/identity/src/main/java/com/adobe/marketing/mobile/identity/ConfigurationSharedStateIdentity.java
```diff
@@ -27,52 +28,35 @@
...
                         sharedState,
                         IdentityConstants.JSON_EXPERIENCE_CLOUD_SERVER_KEY,
                         Defaults.SERVER);
 
-        if (StringUtils.isNullOrEmpty(this.marketingCloudServer)) {
+        if (StringUtils.isNullOrEmpty(server)) {
             this.marketingCloudServer = Defaults.SERVER;
+        } else {
+            this.marketingCloudServer = server;
         }
```
emdobrin, line 60: nit: can use one-liner

yangyansong-adbe (approved):

PR #426, opened 2023-04-06
Event history fixes

rymorale:
## Description
- fix searchType logic within the HistoricalEventsQuerying. if we have search type == any, we don't want to do an ordered search.
- fix fnv1a hash creation within MapExtensions. we want to check for empty map values (null or empty string) and not include them when creating the hash.

## Types of changes

## Checklist:

### DIFF
--- a/code/core/src/main/java/com/adobe/marketing/mobile/internal/util/MapExtensions.kt
+++ b/code/core/src/main/java/com/adobe/marketing/mobile/internal/util/MapExtensions.kt
@@ -22,30 +22,37 @@ import org.json.JSONObject
  */
 @JvmSynthetic
 internal fun Map<String, Any?>.fnv1a32(masks: Array<String>? = null): Long {
-    val flattenedMap = this.flattening()
+    // remove null or empty values before flattening
+    val flattenedMap = this.filterNullOrEmptyValues().flattening()
     val kvPairs = StringBuilder()
     var innerMasks = masks
     if (innerMasks?.isEmpty() == true) innerMasks = null
     innerMasks?.let {
         it.sortedArray().forEach { mask ->
-            // we want to ignore null values
             if (mask.isNotEmpty() && flattenedMap[mask] != null) {
-                val mapValue = flattenedMap[mask].toString()
-                // we want to ignore empty strings
-                if (mapValue != "") {
-                    kvPairs.append(mask).append(":").append(mapValue)
-                }
+                kvPairs.append(mask).append(":").append(flattenedMap[mask].toString())
             }
         }
     } ?: run {
         flattenedMap.toSortedMap().forEach { entry ->
-            // we want to ignore null values or empty strings
-            if (entry.value != null && entry.value.toString() != "") {
-                kvPairs.append(entry.key).append(":").append(entry.value.toString())
-            }
+            kvPairs.append(entry.key).append(":").append(entry.value.toString())
         }
     }
     return StringEncoder.convertStringToDecimalHash(kvPairs.toString())
+}
+
+/**
+ * Removes null or empty string values from a map.
+ *
+ * @return the [Map] with null and empty string values removed
+ */
+@JvmSynthetic
+internal fun Map<String, Any?>.filterNullOrEmptyValues(): Map<String, Any?> {
+    val filteredMap = HashMap<String, Any>()
+    for ((key, value) in this) {
+        if (value != null && value != "") filteredMap[key] = value
+    }
+    return filteredMap
 }
 
 /**
--- a/code/core/src/main/java/com/adobe/marketing/mobile/internal/util/MapExtensions.kt
+++ b/code/core/src/main/java/com/adobe/marketing/mobile/internal/util/MapExtensions.kt
@@ -12,6 +12,7 @@
 package com.adobe.marketing.mobile.internal.util
 
 import com.adobe.marketing.mobile.internal.util.UrlEncoder.urlEncode
+import com.adobe.marketing.mobile.util.StringUtils
 import org.json.JSONObject
 
 /**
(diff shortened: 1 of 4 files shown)


code/core/src/main/java/com/adobe/marketing/mobile/internal/util/MapExtensions.kt
```diff
@@ -28,8 +28,10 @@ internal fun Map<String, Any?>.fnv1a32(masks: Array<String>? = null): Long {
     if (innerMasks?.isEmpty() == true) innerMasks = null
     innerMasks?.let {
         it.sortedArray().forEach { mask ->
-            if (mask.isNotEmpty() && flattenedMap.containsKey(mask)) {
-                kvPairs.append(mask).append(":").append(flattenedMap[mask].toString())
+            val mapValue = flattenedMap[mask].toString()
+            // we want to ignore null or empty values within the map
+            if (mask.isNotEmpty() && mapValue != "null" && mapValue != "") {
```
praveek, line 33: Can you check for null before converting it to String and add some unit test case to cover this case?

code/core/src/main/java/com/adobe/marketing/mobile/launch/rulesengine/HistoricalEventsQuerying.kt
```diff
@@ -32,7 +32,7 @@ internal fun historicalEventsQuerying(
         var eventCounts = 0
         extensionApi.getHistoricalEvents(
             requests.toTypedArray(),
-            searchType == SEARCH_TYPE_ANY
+            searchType != SEARCH_TYPE_ANY
```
praveek, line 35: nit: To future proof, can you define a new constant and change it to searchType == SEARCH_TYPE_ORDERED.

praveek: Can you also add a test to test this case?

rymorale: will get these updated

code/core/src/test/java/com/adobe/marketing/mobile/internal/util/MapUtilsTests.java
```diff
@@ -287,6 +287,92 @@ public void testGetFnv1aHash_WithMaskMatchingNoKeys() {
...
+        // test
+        final long hash =
+                MapUtilsKt.convertMapToFnv1aHash(map, new String[] {"inner.a", "inner.b"});
+        // verify flattened map string "inner.a:1inner.b:2"
+        final long expectedHash = 3328417429L;
+        assertEquals(expectedHash, hash);
+    }
+
+    @Test
+    public void testGetFnv1aHash_NullAndEmptyMapValuesPresent_WithNoMask() {
```
sbenedicadb, line 356: looks good.  given the newly introduced method acts recursively, let's make sure we add a test which covers the recursive behavior as well

code/core/src/main/java/com/adobe/marketing/mobile/launch/rulesengine/HistoricalEventsQuerying.kt
```diff
@@ -32,7 +32,7 @@ internal fun historicalEventsQuerying(
         var eventCounts = 0
         extensionApi.getHistoricalEvents(
             requests.toTypedArray(),
-            searchType == SEARCH_TYPE_ANY
+            searchType == SEARCH_TYPE_ORDERED
```
sbenedicadb, line 35: why are we changing the default search type?

sbenedicadb (changes requested):

code/core/src/main/java/com/adobe/marketing/mobile/launch/rulesengine/HistoricalEventsQuerying.kt
sbenedicadb: ah, nevermind.  this is a bug and you fixed it.  good catch.

code/core/src/main/java/com/adobe/marketing/mobile/internal/util/MapExtensions.kt
```diff
@@ -22,13 +22,14 @@ import org.json.JSONObject
...
-    val flattenedMap = this.flattening()
+    // remove null or empty values before flattening
+    val flattenedMap = this.filterNullOrEmptyValues().flattening()
     val kvPairs = StringBuilder()
     var innerMasks = masks
     if (innerMasks?.isEmpty() == true) innerMasks = null
     innerMasks?.let {
         it.sortedArray().forEach { mask ->
-            if (mask.isNotEmpty() && flattenedMap.containsKey(mask)) {
+            if (mask.isNotEmpty() && flattenedMap[mask] != null) {
```
praveek, line 32: if (mask.isNotEmpty() && !flattenedMap[mask].isNullOrEmptyString())

code/core/src/main/java/com/adobe/marketing/mobile/internal/util/MapExtensions.kt
```diff
@@ -22,13 +22,14 @@ import org.json.JSONObject
...
                 kvPairs.append(mask).append(":").append(flattenedMap[mask].toString())
             }
         }
```
praveek, line 35:
```run {
        flattenedMap.toSortedMap().forEach { entry ->
            if (!entry.value.isNullOrEmptyString()) {
                kvPairs.append(entry.key).append(":").append(entry.value.toString())
            }
        }
    }
```

praveek (approved):

PR #440, opened 2023-04-22
Fix iam scrolling behavior and ui takeover opacity

rymorale:
## Description
- fix for MOB-18539 In-App Message rendering issue on small screens on Android
- improvements for ui takeover opacity
- remove MessageWebView and use androidx CardView to handle rounded corners on the in-app messages.

## Types of changes

## Checklist:


code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/MessageWebView.java
```diff
@@ -1,116 +0,0 @@
...
-import android.graphics.Shader;
-import android.webkit.WebView;
-import java.util.HashMap;
-import java.util.Map;
-
-/**
- * An extension of {@link WebView} that customizes the drawing of the {@link Canvas} to allow for
- * rounded corners.
- */
-class MessageWebView extends WebView {
```
sbenedicadb, line 32: did we check to make sure none of the other extensions aren't using `MessageWebView`?  acs or assurance?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/MessageFragment.java
```diff
@@ -91,10 +91,14 @@ public boolean onTouch(final View view, final MotionEvent motionEvent) {
 
         // determine if the tapped view is the webview
         if (view.getId() == message.webView.getId()) {
-            // pass the event to the gesture detector to determine if a motion event occurred on the
-            // webview.
-            gestureDetector.onTouchEvent(motionEvent);
-            // perform the tap to allow interaction with buttons within the webview
+            // if we have no gestures just pass the touch event to the webview
+            if (!message.getMessageSettings().getGestures().isEmpty()) {
```
sbenedicadb, line 95: two things here.  

- do we know that gestures returned from `getGestures` is never going to be null?  might be good to do a null check.  
- ideally the code matches the comments (comment says if no gestures then pass through, but the condition checks for the presence of gestures then implements that case).  here it's probably easier to read if you have the early exit (no gestures) at the top of the method, then the rest of the method doesn't need to be wrapped in the `if` block.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/MessageWebViewRunner.java
```diff
@@ -295,26 +307,44 @@ private void createMessageFrameAndAddMessageToRootView(final MessageSettings set
...
+                                    .getAppContextService()
+                                    .getApplicationContext());
+            backdrop.setLayoutParams(
+                    new LayoutParams(LayoutParams.MATCH_PARENT, LayoutParams.MATCH_PARENT));
+            backdrop.setBackgroundColor(Color.parseColor(settings.getBackdropColor()));
+
+            // alpha values range from 0-255 (0 means fully transparent, 255 means fully opaque)
+            final int convertedAlpha =
+                    (int) (settings.getBackdropOpacity() * FULLY_OPAQUE_ALPHA_VALUE);
+            backdrop.getBackground().setAlpha(convertedAlpha);
```
sbenedicadb, line 337: 👍

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/MessageWebView.java
rymorale: ACS and Target both use the AEPMessage class for displaying things in a webview. The webview is created internally within the AEPMessage / MessageFragment class though so removing the MessageWebView and using the standard WebView shouldn't cause dependency issues. I will setup tests to verify this.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/MessageWebViewRunner.java
```diff
@@ -295,26 +307,44 @@ private void createMessageFrameAndAddMessageToRootView(final MessageSettings set
...
+        final Context context =
+                ServiceProvider.getInstance().getAppContextService().getApplicationContext();
+        webViewFrame = new CardView(context);
+        final float calculatedRadius =
+                TypedValue.applyDimension(
+                        TypedValue.COMPLEX_UNIT_DIP,
+                        settings.getCornerRadius(),
+                        context.getResources().getDisplayMetrics());
+        webViewFrame.setRadius(calculatedRadius);
+        webViewFrame.addView(message.webView);
```
prudrabhat, line 320: Is CardView being used only for the ability to set the radius? If so, have you already tried to change the shape of the layout programmatically applying a shape(something like [this](https://github.com/adobe/aepsdk-assurance-android/blob/dev/code/assurance/src/main/res/drawable/shape_custom_button_filled.xml#L16) but in code) ? I ask because, adding the CardView dependency will require implementers that do not need Messaging to pull this dependency. 

If there is no other way to apply this rounded corners, then we should add this to the [implicit dependency in pom as well](https://github.com/adobe/aepsdk-core-android/blob/main/code/core/release.gradle#L51)

rymorale: I did try applying a rounded rectangle as a background for the webview but it was not working as desired as the webview would retain the shape specified in the html content.

prudrabhat: Just to clarify - did you also try wrapping the webview inside a FrameLayout and then applying the shape to the Frame layout?

code/core/release.gradle
```diff
@@ -51,10 +51,20 @@ publishing {
                 withXml {
                     def dependenciesNode = asNode().appendNode('dependencies')
 
-                    def coreDependencyNode = dependenciesNode.appendNode('dependency')
-                    coreDependencyNode.appendNode('groupId', 'org.jetbrains.kotlin')
-                    coreDependencyNode.appendNode('artifactId', 'kotlin-stdlib-jdk8')
-                    coreDependencyNode.appendNode('version', rootProject.kotlin_version)
+                    def kotlinDependencyNode = dependenciesNode.appendNode('kotlin')
```
praveek, line 54: def kotlinDependencyNode = dependenciesNode.appendNode('dependency')

code/core/release.gradle
```diff
@@ -51,10 +51,20 @@ publishing {
...
+                    kotlinDependencyNode.appendNode('groupId', 'org.jetbrains.kotlin')
+                    kotlinDependencyNode.appendNode('artifactId', 'kotlin-stdlib-jdk8')
+                    kotlinDependencyNode.appendNode('version', rootProject.kotlin_version)
+
+                    def cardviewDependencyNode = dependenciesNode.appendNode('cardview')
```
praveek, line 59: def cardviewDependencyNode = dependenciesNode.appendNode('dependency')

code/core/release.gradle
```diff
@@ -51,10 +51,20 @@ publishing {
...
+                    cardviewDependencyNode.appendNode('groupId', 'androidx.cardview')
+                    cardviewDependencyNode.appendNode('artifactId', 'cardview')
+                    cardviewDependencyNode.appendNode('version', '1.0.0')
+
+                    def appcompatDependencyNode = dependenciesNode.appendNode('appcompat')
```
praveek, line 64: def appcompatDependencyNode = dependenciesNode.appendNode('dependency')

rymorale: whoops, fixed this for all 3 dependencies

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/MessageWebViewRunner.java
rymorale: i was applying the shape as a background to the webview. i just tried applying the shape as a background to a FrameLayout then added the webview as a child view which then gave the same incorrect behavior as I saw before.

prudrabhat: Got it. Thank you for testing that Ryan! We recently had some requests around removing view dependencies as much as possible in core/assurance. So just wanted to make sure we explored alternatives if they exists.

sbenedicadb (approved):

sbenedicadb (approved):

PR #458, opened 2023-05-11
Add interfaces to be used by different displayable ui elements / messages

rymorale:
Add Showable, Dismissible, and Presentable interfaces which displayable UI Elements like in-app messages and alert messages will implement. This will allow for these UI elements to be managed by the same delegate.
## Types of changes

## Checklist:

### DIFF
--- a/code/core/src/main/java/com/adobe/marketing/mobile/services/MessagingDelegate.java
+++ b/code/core/src/main/java/com/adobe/marketing/mobile/services/MessagingDelegate.java
@@ -12,6 +12,7 @@
 package com.adobe.marketing.mobile.services;
 
 import com.adobe.marketing.mobile.services.ui.FullscreenMessage;
+import com.adobe.marketing.mobile.services.ui.Showable;
 
 /**
  * UI Message delegate which is used to listen for current message lifecycle events and control if
@@ -23,8 +24,18 @@ public interface MessagingDelegate {
      *
      * @param message {@link FullscreenMessage} that is being displayed
      */
+    @Deprecated
     default void onShow(final FullscreenMessage message) {
         Log.debug(ServiceConstants.LOG_TAG, "MessagingDelegate", "Fullscreen message shown.");
+    }
+
+    /**
+     * Invoked when a {@code Showable} message is displayed.
+     *
+     * @param message {@link Showable} message that is being displayed
+     */
+    default void onShow(final Showable message) {
+        Log.debug(ServiceConstants.LOG_TAG, "MessagingDelegate", "Message shown.");
     }
 
     /**
@@ -32,8 +43,18 @@ public interface MessagingDelegate {
      *
      * @param message {@link FullscreenMessage} that is being dismissed
      */
+    @Deprecated
     default void onDismiss(final FullscreenMessage message) {
         Log.debug(ServiceConstants.LOG_TAG, "MessagingDelegate", "Fullscreen message dismissed.");
+    }
+
+    /**
+     * Invoked when a {@code Showable} message is dismissed.
+     *
+     * @param message {@link Showable} message that is being dismissed
+     */
+    default void onDismiss(final Showable message) {
+        Log.debug(ServiceConstants.LOG_TAG, "MessagingDelegate", "Message shown.");
     }
 
     /**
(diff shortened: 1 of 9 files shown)


code/core/src/main/java/com/adobe/marketing/mobile/services/ui/Presentable.java
```diff
@@ -0,0 +1,15 @@
...
+  Unless required by applicable law or agreed to in writing, software distributed under
+  the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+  OF ANY KIND, either express or implied. See the License for the specific language
+  governing permissions and limitations under the License.
+*/
+
+package com.adobe.marketing.mobile.services.ui;
+
+/** Represents a UI element which is both showable and dismissible */
+public interface Presentable extends Showable, Dismissible {}
```
prudrabhat, line 15: Are there any components that would implement show but not dismiss (or the other way around). More specifically, do we anticipate components that can only be shown and never dismissed? Asking because I am unclear on why `Showable` and `Dismissible` cannot be eliminated in favor of just having the `show()` and `dismiss()` in `Presentable` directly

code/core/src/main/java/com/adobe/marketing/mobile/services/MessagingDelegate.java
```diff
@@ -23,38 +24,80 @@ public interface MessagingDelegate {
...
+    @Deprecated
     boolean shouldShowMessage(final FullscreenMessage message);
 
+    /**
+     * Used to determine if a message should be shown.
+     *
+     * @param message {@link Showable} message that is about to get displayed
+     * @return true if the message should be displayed, false otherwise
+     */
+    default boolean shouldShowMessage(final Showable message) {
```
prudrabhat, line 75: Does this apply only to messages? Given that this interface now handles everything in terms of presentables/showables - what are your thoughts on naming this to be more generic (.. `shouldShow(...)`)  ?

code/core/src/main/java/com/adobe/marketing/mobile/services/ui/Presentable.java
rymorale: I am open to making that change. In this pr I was updating android to match the ios interfaces created.

code/core/src/main/java/com/adobe/marketing/mobile/services/MessagingDelegate.java
rymorale: i considered changing the naming of the function but the function itself is within a `MessagingDelegate` interface. I feel like `showShowMessage` fits better?

code/core/src/main/java/com/adobe/marketing/mobile/services/MessagingDelegate.java
```diff
@@ -23,38 +24,80 @@ public interface MessagingDelegate {
...
     default void onShow(final FullscreenMessage message) {
         Log.debug(ServiceConstants.LOG_TAG, "MessagingDelegate", "Fullscreen message shown.");
     }
 
+    /**
+     * Invoked when a {@code Showable} message is displayed.
+     *
+     * @param message {@link Showable} message that is being displayed
+     */
+    default void onShow(final Showable message) {
```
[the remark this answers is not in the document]
rymorale: similar to the presentable interface below, i was matching the parameters of the ios messaging delegate: https://github.com/adobe/aepsdk-core-ios/blob/main/AEPServices/Sources/ui/MessagingDelegate.swift.

code/core/src/main/java/com/adobe/marketing/mobile/services/MessagingDelegate.java
prudrabhat: 👍🏼 Got it. I can go either way.

code/core/src/main/java/com/adobe/marketing/mobile/services/MessagingDelegate.java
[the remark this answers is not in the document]
rymorale: I had a local change which I had mistakenly forgot to add to this pr. I had added a `getParent()` function in the Presentable interface. I'll get it merged now.

PR #461, opened 2023-05-17
In-app message fixes

rymorale:
## Types of changes

## Checklist:

### DIFF
--- a/code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/AEPMessage.java
+++ b/code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/AEPMessage.java
@@ -64,7 +64,6 @@ class AEPMessage implements FullscreenMessage {
     int parentViewWidth;
     final FullscreenMessageDelegate listener;
     final MessagesMonitor messagesMonitor;
-    MessageFragment messageFragment;
 
     // private vars
     private WebView webView;
@@ -77,6 +76,7 @@ class AEPMessage implements FullscreenMessage {
     private Map<String, String> assetMap = Collections.emptyMap();
     private final Executor executor;
     private MessageWebViewClient webViewClient;
+    private MessageFragment messageFragment;
 
     /**
      * Constructor.
@@ -234,11 +234,21 @@ class AEPMessage implements FullscreenMessage {
     /** Dismisses the message. */
     @Override
     public void dismiss() {
+        dismiss(false);
+    }
+
+    /**
+     * Dismisses the message.
+     *
+     * @param dismissedWithBackTouch {@code boolean} signaling if the dismiss was triggered by a
+     *     back button press
+     */
+    public void dismiss(final boolean dismissedWithBackTouch) {
         if (!messagesMonitor.dismiss()) {
             return;
         }
         // add a dismiss animation if the webview wasn't previously removed via a swipe gesture
-        if (!messageFragment.dismissedWithGesture) {
+        if (!messageFragment.isDismissedWithGesture()) {
             dismissAnimation = setupDismissAnimation();
             animationListener =
                     new Animation.AnimationListener() {
@@ -248,7 +258,7 @@ class AEPMessage implements FullscreenMessage {
                         @Override
                         public void onAnimationEnd(final Animation animation) {
                             // wait for the animation to end then clean the views
-                            cleanup();
+                            cleanup(dismissedWithBackTouch);
                         }
 
                         @Override
@@ -257,9 +267,10 @@ class AEPMessage implements FullscreenMessage {
             dismissAnimation.setAnimationListener(animationListener);
             webViewFrame.startAnimation(dismissAnimation);
             return;
-        } // otherwise, just clean the views
+        }
 
-        cleanup();
+        // otherwise, just clean the views
+        cleanup(dismissedWithBackTouch);
     }
 
     /**
@@ -330,11 +341,20 @@ class AEPMessage implements FullscreenMessage {
         return html;
     }
 
-    /** Tears down views and listeners used to display the {@link AEPMessage}. */
-    void cleanup() {
+    /**
+     * Tears down views and listeners used to display the {@link AEPMessage}.
+     *
+     * @param dismissedWithBackTouch {@code boolean} signaling if the dismiss was triggered by a
+     *     back button press
+     */
+    void cleanup(final boolean dismissedWithBackTouch) {
         Log.trace(ServiceConstants.LOG_TAG, TAG, "Cleaning the AEPMessage.");
 
-        listener.onDismiss(this);
+        if (dismissedWithBackTouch) {
+            listener.onBackPressed(this);
+        } else {
+            listener.onDismiss(this);
+        }
         webViewFrame.setOnTouchListener(null);
         webView.setOnTouchListener(null);
         if (dismissAnimation != null) {
@@ -393,7 +413,7 @@ class AEPMessage implements FullscreenMessage {
 
     /**
      * Recreates the {@link WebView} frame used for displaying the Messaging IAM using the {@link
-     * MessageWebViewRunner}. This method should be called after a device orientation change occurs.
+     * MessageWebViewUtil}. This method should be called after a device orientation change occurs.
      *
      * @param parentViewWidth {@code int} containing the width of the parent activity
      * @param parentViewHeight {@code int} containing the height of the parent activity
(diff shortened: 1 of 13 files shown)


prudrabhat:
Hey @rymorale - would you be willing to add details about the list of changes being made and the reasons they are needed in the PR description. I feel that the context, will be helpful for review.

rymorale:
> Hey @rymorale - would you be willing to add details about the list of changes being made and the reasons they are needed in the PR description. I feel that the context, will be helpful for review.

Sure, it looks like the issues fixed got buried in the first commit's description.  The issues are:
- fix iam renderring when device orientation changed
- fix back button detection for iam dismissal
- notify message delegate when url is loaded
- fix iam cardview transparency issues

prudrabhat (commented):
I have been ignorant of the UI services part of core as of now and have only started to look at it about a week back 😅 . So, my understanding and the following comment likely has gaps, but here are couple thoughts I have -

- deep circular inter dependancies among `AEPMessage`, `MessageFragment`, `Message` can be avoided if (in addition to a few modifications) the creation of the view related elements of the AEPMessage are done during construction and not during `show()`.  I am curious to know if there is any specific reason we do it this way. Are all the instances of AEPMessages created at once by the Messaging extension and displayed later? Or have we seen performance hits when we eager initialize and attach at a later time?
- Based on my understanding, the view hierarchy is - _MessageFragment_ -> _WebView frame_ -> _WebView_ . In code,  `AEPMessage` holds and instance of  WebView, the CardView composing the WebView and the `MessageFragment` that is the slot where the entire in app message is displayed. I think the relationship between the MessageFragment and the AEPMessage is inverted, causing the juggling of webview among the classes to set dimensions/listeners in different places. I am wondering if things can be simplified if there exists an orchestrating layer that manages content in the container/fragment and all the requests to show and dismiss messages go through it. We can discuss more if you find this promising.

I am not proposing that we look in to the possibility of making these change this in this PR, but I think that modifying the FullscreenMessage parts of this code is going to become tricky further down the lane unless we invest in ironing out the convoluted flow. 

Lastly, would you be willing  to profile this change with an IAM simulation and see if there are any orphaned objects retained. I ask because, @PravinPK and I profiled a scenario which seem to be leading to AEPMessage objects not being GC'd  most likely due to listeners.

prudrabhat (commented):
I have been ignorant of the UI services part of the core as of now and have only started to look at it about a week back 😅 . So, my understanding and the following comment likely has gaps, but here are couple thoughts I have -

- deep circular inter dependancies among `AEPMessage`, `MessageFragment`, `Message` can be avoided if (in addition to a few modifications) the creation of the view related elements of the AEPMessage are done during construction and not during `show()`.  I am curious to know if there is any specific reason we do it this way. Are all the instances of AEPMessages created at once by the Messaging extension and displayed later? Or have we seen performance hits when we eager initialize and attach at a later time?
- Based on my understanding, the view hierarchy is - _MessageFragment_ -> _WebView frame_ -> _WebView_ . In code,  `AEPMessage` holds and instance of  WebView, the CardView composing the WebView and the `MessageFragment` that is the slot where the entire in app message is displayed. I think the relationship between the MessageFragment and the AEPMessage is inverted, causing the juggling of webview among the classes to set dimensions/listeners in different places. I am wondering if things can be simplified if there exists an orchestrating layer that manages content in the container/fragment and all the requests to show and dismiss messages go through it. We can discuss more if you find this promising.

I am not proposing that we look in to the possibility of making these change this in this PR, but I think that modifying the FullscreenMessage parts of this code is going to become tricky further down the lane unless we invest in ironing out the convoluted flow. 

Lastly, would you be willing  to profile this change with an IAM simulation and see if there are any orphaned objects retained. I ask because, @PravinPK and I profiled a scenario which seem to be leading to AEPMessage objects not being GC'd  most likely due to listeners.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/AEPMessage.java
```diff
@@ -275,7 +252,32 @@ public void dismiss() {
         if (!messagesMonitor.dismiss()) {
             return;
         }
-        removeFromRootViewGroup();
+        // add a dismiss animation if the webview wasn't previously removed via a swipe gesture
+        if (!messageFragment.dismissedWithGesture) {
+            dismissAnimation = setupDismissAnimation();
```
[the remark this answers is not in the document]
rymorale: i'll leave it at class level so i can clean the listener when the message is dismissed. or would that no longer be needed if it was a local var?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/AEPMessage.java
```diff
@@ -467,6 +391,29 @@ public void setMessageSetting(final MessageSettings messageSettings) {
...
+     * @param parentViewWidth {@code int} containing the width of the parent activity
+     * @param parentViewHeight {@code int} containing the height of the parent activity
+     */
+    void recreateWebViewFrame(final int parentViewWidth, final int parentViewHeight) {
+        this.parentViewWidth = parentViewWidth;
+        this.parentViewHeight = parentViewHeight;
+
+        try {
+            MessageWebViewRunner messageWebViewRunner = new MessageWebViewRunner(this);
+            messageWebViewRunner.run();
```
prudrabhat, line 407: I believe we can reduce/break one hop of circular dependency of components containing AEPMessage in the following way. The explanation may look long but the change will likely be simple.

`recreateWebViewFrame()` currently is called either from `MessageFragment#onActivityCreated()` and `MessageFragment.layoutChangeListener#onLayoutChange()` - both of which are will be called on the main thread. Even here (rightly), `messageWebViewRunner.run()` is being called directly resulting in the invocation to be on the calling thread (which is always the main thread in this case). 

So, MessageWebViewRunner need not be a runnable any more.  It can just be a utility class with a `show()` with functionality in the run method.  That will allow you would call `messageWebViewRunner.show()` without creating a thread.

Furthermore,  changing this to be a utility class will allow you to pass the `AEPMessage` on demand  to the `show()`  resulting in removing the need to hold a reference to the `AEPMessage` in this class.
The `MessageWebViewRunner` now just has one public method `show(final AEPMessage message)`. 

Now, looking at the constructor, it initializes member variable `settings`, `messageHeight` etc derived from `AEPMessage` itself. I think this can also be done directly just before we call `MessageWebViewRunner#show(AEPMessage)` (formerly run() ).

Given that `MessageWebViewRunner` is stateless we no longer need to create multiple instances of this runner and just use one instance inside AEPMessage on construction. 
Let me know what you think.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/MessageWebViewRunner.java
```diff
@@ -473,17 +353,4 @@ private int getOriginY(final MessageSettings settings) {
...
-    /**
-     * Sets the asset map which contains the mapping between a remote image asset url and it's
-     * cached location.
-     *
-     * @param assetMap The {@code Map<String, String} object containing the mapping between a remote
-     *     asset url and its cached location.
-     */
-    void setLocalAssetsMap(final Map<String, String> assetMap) {
-        if (assetMap != null && !assetMap.isEmpty()) {
-            this.assetMap = new HashMap<>(assetMap);
```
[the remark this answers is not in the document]
rymorale: this functionality has shifted over to the AEPMessage class where it is set within the webview client when creating a webview:
 ```
webViewClient = new MessageWebViewClient(AEPMessage.this);
webViewClient.setLocalAssetsMap(assetMap);
newWebView.setWebViewClient(webViewClient);
```

File: code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/MessageWebViewUtil.java [hunk 4]
```diff
@@ -44,44 +41,35 @@ class MessageWebViewRunner implements Runnable {
...
-     * Validates the message passed in the constructor. If valid, the webview is created and an
+     * Validates the passed in {@code AEPMessage}. If valid, the {@link WebView} is created and an
      * attempt is made to show the message.
+     *
+     * @param message {@link AEPMessage} containing the in-app message payload
      */
     @SuppressWarnings("SetJavaScriptEnabled")
-    @Override
-    public void run() {
+    public void show(final AEPMessage message) {
```
prudrabhat, line 56: nit: This can be package private.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/AEPMessage.java
```diff
@@ -403,8 +423,8 @@ void recreateWebViewFrame(final int parentViewWidth, final int parentViewHeight)
         this.parentViewHeight = parentViewHeight;
 
         try {
-            MessageWebViewRunner messageWebViewRunner = new MessageWebViewRunner(this);
-            messageWebViewRunner.run();
+            MessageWebViewUtil messageWebViewUtil = new MessageWebViewUtil();
```
prudrabhat, line 426: nit: You can instantiate this only once as a `private final` member.

prudrabhat (approved):
Minor suggestions. LGTM otherwise. Recommend doing a final sanity profiling before merging.

prudrabhat (approved):
Minor suggestions. LGTM otherwise. Recommend doing a final sanity profiling before merging.

PR #519, opened 2023-08-07
fix map casting exception when attempting to show a local notification

rymorale:
## Description
A ClassCastException was occurring because the local notification user info map was unmodifiable. To resolve this, a HashMap copy of the local notification user info map should be made instead of casting the existing user info map to a HashMap.
Internal JIRA: MOB-19323

## Types of changes

## Checklist:

### DIFF
--- a/code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/AndroidUIService.java
+++ b/code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/AndroidUIService.java
@@ -246,7 +246,9 @@ public class AndroidUIService implements UIService {
                 !MapUtils.isNullOrEmpty(notificationSetting.getUserInfo())
                         ? new HashMap<>(notificationSetting.getUserInfo())
                         : null;
-        intent.putExtra(NOTIFICATION_USER_INFO_KEY, userInfo);
+        if (!MapUtils.isNullOrEmpty(userInfo)) {
+            intent.putExtra(NOTIFICATION_USER_INFO_KEY, userInfo);
+        }
         intent.putExtra(NOTIFICATION_SOUND_KEY, notificationSetting.getSound());
         intent.putExtra(NOTIFICATION_TITLE, notificationSetting.getTitle());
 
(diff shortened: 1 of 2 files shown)


code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/AndroidUIService.java
```diff
@@ -241,9 +242,11 @@ public void showLocalNotification(final NotificationSetting notificationSetting)
...
         intent.putExtra(NOTIFICATION_DEEPLINK_KEY, notificationSetting.getDeeplink());
         intent.putExtra(NOTIFICATION_CONTENT_KEY, notificationSetting.getContent());
-        intent.putExtra(
-                NOTIFICATION_USER_INFO_KEY,
-                (HashMap<String, Object>) notificationSetting.getUserInfo());
+        final HashMap<String, Object> userInfo =
+                !MapUtils.isNullOrEmpty(notificationSetting.getUserInfo())
+                        ? new HashMap<>(notificationSetting.getUserInfo())
+                        : null;
+        intent.putExtra(NOTIFICATION_USER_INFO_KEY, userInfo);
```
prudrabhat, line 249: Which component creates the `NotificationSettings` in the case of an error? Is this the customer application or the extension?
I ask because the `userInfo` does not seem to have a contract that says the contents are immutable (I understand that they rightly need to be immutable). So I am wondering if this issue can be fixed at the component that creates this setting by using a mutable map instead.

prudrabhat, line 249: nit: you can do away with fetching UserInfo once by 
```
final HashMap<String, Object> userInfo = notificationSetting.getUserInfo();
intent.putExtra(NOTIFICATION_USER_INFO_KEY, (userInfo != null ? new HashMap<>(userInfo) : null));
```

Further, I think `NOTIFICATION_USER_INFO_KEY` can only be added if contents exist instead of adding a null value explicitly.
```
    final HashMap<String, Object> userInfo = notificationSetting.getUserInfo();
    if (!MapUtils.isNullOrEmpty(userInfo)) {
        intent.putExtra(NOTIFICATION_USER_INFO_KEY, new HashMap<>(userInfo));
    }
```

prudrabhat: Based on internal discussion, this may be needed for more than one extension. I am also inclined with adding it in core.

rymorale: `notificationSetting.getUserInfo()` is returning a Map so assigning it to `final HashMap<String, Object> userInfo` will not work without casting (which what the original problematic code was doing). I'll update the new code to only add the NOTIFICATION_USER_INFO_KEY if the user info map is non empty.

prudrabhat (approved):

PR #554, opened 2023-09-18
Add PresentationObserver for presentation co-ordination

prudrabhat:
## Description
This PR adds support for managing presentation conflicts when multiple presentations are visible or are being requested to be shown. This is similar to MessagingMonitor in the existing UIService but the difference is that it handles all presentations managed by new UIService.

- Added `PresentationObserver` that is capable of holding information about presentations when they become visible and invisible from the screen. This is currently local to `AEPPresentable`. When need arises it can be moved to the API layer.
- Added `AEPPresentable#hasConflicts` as mechanism for all presentables to evaluate if the existing visible presentations are conflicting for showing the current presentation. The result of this evaluation will determine whether the show request will be processed further.
- Added unit tests

## Related Issue
#530 

## Motivation and Context
Support for managing presentation conflicts when multiple presentations are visible or are being requested to be shown is needed to ensure that Alert, InAppMessage ui do not overlap etc 

## How Has This Been Tested?
- Unit tests
- Manual tests with test app

## Types of changes

## Checklist:


code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/vnext/alert/AlertPresentable.kt
```diff
@@ -67,4 +69,9 @@ internal class AlertPresentable(
     override fun getPresentation(): Alert {
         return alert
     }
+
+    override fun hasConflicts(visiblePresentations: List<Presentation<*>>): Boolean {
+        // Only show if there are no other alerts or in-app messages visible
+        return visiblePresentations.any { (it is Alert || it is InAppMessage) }
```
sbenedicadb, line 75: this determination is actually driven by the `mobileParameters` setting `uiTakeover`. wondering if there's a way to just check for that value, as it also sets us up to automatically handle future "ui takeover" messages as well

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/vnext/message/InAppMessagePresentable.kt
```diff
@@ -127,6 +129,11 @@ internal class InAppMessagePresentable(
         // now wait for onDisposed to be called on the composable
     }
 
+    override fun hasConflicts(visiblePresentations: List<Presentation<*>>): Boolean {
+        // InAppMessages can be shown if there are no other visible in-app messages or alerts
+        return visiblePresentations.any { (it is InAppMessage || it is Alert) }
```
sbenedicadb, line 134: same as the comment above for alerts

sbenedicadb (approved):
looks good to me - just one question about `hasConflicts` implementations we can discuss

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/vnext/alert/AlertPresentable.kt
prudrabhat: The `visiblePresentations` does hold all `Presentation` objects which have `PresentationSettings` . So we _can_ can access the `uiTakeover` settings here.
However, looking at the [current code in Core 2.0](https://github.com/search?q=repo%3Aadobe%2Faepsdk-core-android%20uiTakeover&type=code), I see that `uiTakeover` is being used only during determination of whether the message is to be dismissed due to a background tap/back button press  and the decision to show is being done based on presentation delegate + [current state of the tree as seen by `MessagesMonitor`](https://github.com/adobe/aepsdk-core-android/blob/main/code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/AEPMessage.java#L219). Is this a bug in 2.0 😅  ?

prudrabhat: Talked with @sbenedicadb on this topic offline. Learned that there have been discussions about allowing multiple UI's to be displayed on the screen simultaneously depending on feasibility (ex: a center aligned modal InAppMessage can co-exist with a banner style InAppMessage when they do not take over/intercept all ui operations - `uiTakeOver = false` ). We discussed that the current implementation of `hadConflicts` merely checking the type of visible presentations will not be sufficient to achieve the behavior.  
We agreed that the current design, however, allows multiple presentations to co-ordinate based on the `PresentationSettings` for each presentation and this can be added later when the feature is needed.

praveek (approved):

PR #569, opened 2023-10-09
Allow PresentationUtilityProvider.getApplication() to return nulls

prudrabhat:
## Description
UIService requires `PresentationUtilityProvider` to provide a valid application instance. Intention is for extensions to call `ServiceProvider.appContextService.getApplication()`  but ServiceProvider.appContextService.getApplication() is nullable and can return a null value if the implementer does not set the application via `MobileCore.setApplication()`. In such cases implementers of the `PresentationUtilityProvider` interface as it is currently implemented are forced to return a non-null value but cannot guarantee that. Circumvent this scenario by allowing nulls in `PresentationUtilityProvider.getApplication()` and throwing an exception in such situations without proceeding further.

## Related Issue
N/A

## Motivation and Context
UIService requires `PresentationUtilityProvider` to provide a valid application instance. Intention is for extensions to call `ServiceProvider.appContextService.getApplication()`  but ServiceProvider.appContextService.getApplication() is nullable and can return a null value if the implementer does not set the application via `MobileCore.setApplication()`. In such cases implementers of the `PresentationUtilityProvider` interface as it is currently implemented are forced to return a non-null value but cannot guarantee that. Circumvent this scenario by allowing nulls in `PresentationUtilityProvider.getApplication()` and throwing an exception in such situations without proceeding further.

## How Has This Been Tested?
- Unit tests

## Types of changes

## Checklist:

### DIFF
--- a/code/core/src/test/java/com/adobe/marketing/mobile/services/ui/AEPUIServiceTest.kt
+++ b/code/core/src/test/java/com/adobe/marketing/mobile/services/ui/AEPUIServiceTest.kt
@@ -31,6 +31,7 @@ import org.mockito.MockitoAnnotations
 import kotlin.test.assertEquals
 import kotlin.test.assertNotNull
 import kotlin.test.assertTrue
+import kotlin.test.fail
 
 class AEPUIServiceTest {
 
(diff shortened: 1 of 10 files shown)


code/core/src/test/java/com/adobe/marketing/mobile/services/ui/AEPUIServiceTest.kt
```diff
@@ -96,4 +97,20 @@ class AEPUIServiceTest {
...
+
+    @Test
+    fun `Test that #create throws an exception when application is null`() {
+        val iamSettings: InAppMessageSettings = InAppMessageSettings.Builder().build()
+        val iamEventListener: InAppMessageEventListener = mock(InAppMessageEventListener::class.java)
+        val iamPresentation = InAppMessage(iamSettings, iamEventListener)
+
+        `when`(mockPresentationUtilityProvider.getApplication()).thenReturn(null)
+
+        try {
```
praveek, line 109: Can we use `assertThrows` instead?

prudrabhat: Ah. My bad. Missed this pointer. Will address as a followup/

Issue #311, opened 2023-01-03
Improve release workflows


prudrabhat:
Currently releasing core/identity/lifecycle/signal to maven or maven staging requires a manual trigger from the Github Actions dashboard. In addition to this separate workflow files for main and staging.
- Investigate if the release can be automated for the sub-projects where changes have been made.
- Attempt to unify the actions for stage and main.

PR #592, opened 2023-12-01
Update GH Action and scripts

praveek:
Fixes https://github.com/adobe/aepsdk-core-android/issues/311

Remove hardcoded version numbers from tests.
Cleaned up `versions.sh` script for updating and verifying versions.
Modify the Release GitHub Action to create tags and verify versions.


scripts/version.sh
```diff
@@ -0,0 +1,112 @@
...
+    update 
+else 
+    verify
+fi
+echo "$LINE"
+
+
+
+
+
```
prudrabhat, line 112: nit: new extra lines breaks.

.github/workflows/maven-release.yml
```diff
@@ -2,35 +2,32 @@ name: Publish Release
...
+        - signal
+      
+      tag:
+        description: 'tag/version'
+        required: true        
+  
+      action_tag:
+        description: 'Create tag? ("no" to skip)'
+        required: true
+        default: 'yes'
```
prudrabhat, line 21: How does the `boolean` input look like on the Workflow UI? Wondering if we can replace this with a boolean input if feasible.

prudrabhat (approved):

Issue #599, opened 2024-01-05
Extend rulesengine to support token replacement in arrays


praveek:
The current implementation of the Rules Engine only supports token replacement in Strings or Maps. Extend it to support arrays

PR #600, opened 2024-01-19
[PLATIR-35300] Add array replace token support

yangyansong-adbe:
Fixes https://github.com/adobe/aepsdk-core-android/issues/599

Ported the iOS implementation to Android https://github.com/adobe/aepsdk-core-ios/pull/988

### DIFF
--- a/code/core/src/main/java/com/adobe/marketing/mobile/launch/rulesengine/LaunchRulesConsequence.kt
+++ b/code/core/src/main/java/com/adobe/marketing/mobile/launch/rulesengine/LaunchRulesConsequence.kt
@@ -151,24 +151,29 @@ internal class LaunchRulesConsequence(
         return RuleConsequence(consequence.id, consequence.type, tokenReplacedMap)
     }
 
+    private fun replaceToken(value: Any?, tokenFinder: TokenFinder): Any? {
+        return when (value) {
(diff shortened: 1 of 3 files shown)


code/core/src/main/java/com/adobe/marketing/mobile/launch/rulesengine/LaunchRulesConsequence.kt
```diff
@@ -151,24 +151,29 @@ internal class LaunchRulesConsequence(
         return RuleConsequence(consequence.id, consequence.type, tokenReplacedMap)
     }
 
+    private fun replaceToken(value: Any?, tokenFinder: TokenFinder): Any? {
+        return when (value) {
+            is String -> replaceToken(value, tokenFinder)
+            is Map<*, *> -> replaceToken(EventDataUtils.castFromGenericType(value), tokenFinder)
+            is List<*> -> replaceToken(value, tokenFinder)
+            else -> null
```
praveek, line 159: This should return the original value instead of null.

praveek: Can you also add some tests for this scenario?

praveek (changes requested):

praveek (approved):

PR #612, opened 2024-02-23
[NetworkService] check Internet availability before sending requests

yangyansong-adbe:
## Types of changes

## Checklist:

### DIFF
--- a/code/core/build.gradle.kts
+++ b/code/core/build.gradle.kts
@@ -44,6 +44,7 @@ apiValidation {
     ))
 }
 
+//TODO: Consider moving this to the aep-library plugin later
 dependencies {
     androidTestImplementation("org.mockito:mockito-core:5.3.1")
     androidTestImplementation("com.linkedin.dexmaker:dexmaker-mockito-inline:2.28.3")
--- a/code/core/build.gradle.kts
+++ b/code/core/build.gradle.kts
@@ -43,3 +43,8 @@ apiValidation {
             "com.adobe.marketing.mobile.core.BuildConfig"
     ))
 }
+
+dependencies {
+    androidTestImplementation("org.mockito:mockito-core:5.3.1")
+    androidTestImplementation("com.linkedin.dexmaker:dexmaker-mockito-inline:2.28.3")
+}
--- a/code/core/src/phone/java/com/adobe/marketing/mobile/internal/util/NetworkUtils.kt
+++ b/code/core/src/phone/java/com/adobe/marketing/mobile/internal/util/NetworkUtils.kt
@@ -1,0 +1,34 @@
+/*
+  Copyright 2024 Adobe. All rights reserved.
+  This file is licensed to you under the Apache License, Version 2.0 (the "License");
+  you may not use this file except in compliance with the License. You may obtain a copy
+  of the License at http://www.apache.org/licenses/LICENSE-2.0
+  Unless required by applicable law or agreed to in writing, software distributed under
+  the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+  OF ANY KIND, either express or implied. See the License for the specific language
+  governing permissions and limitations under the License.
+*/
+
+package com.adobe.marketing.mobile.internal.util
+
+import android.net.ConnectivityManager
+import android.net.NetworkCapabilities
+import android.os.Build
+
+/**
+ * Checks if the network is configured to reach the general Internet.
+ *
+ * @param connectivityManager the [ConnectivityManager] to use to check the network status
+ * @return `true` if the network is configured to reach the general Internet, `false` otherwise
+ */
+fun isInternetAvailable(connectivityManager: ConnectivityManager): Boolean {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+        // The getActiveNetwork() API was introduced in API version 23.
+        val network = connectivityManager.activeNetwork
+        val activeCapabilities = connectivityManager.getNetworkCapabilities(network) ?: return false
+        return activeCapabilities.hasCapability(NetworkCapabilities.NET_CAPABILITY_INTERNET)
+    } else {
+        val networkInfo = connectivityManager.activeNetworkInfo ?: return false
+        return networkInfo.isConnected
+    }
+}
--- a/code/core/src/phone/java/com/adobe/marketing/mobile/internal/util/NetworkUtils.kt
+++ b/code/core/src/phone/java/com/adobe/marketing/mobile/internal/util/NetworkUtils.kt
@@ -26,7 +26,7 @@ import android.os.Build
 internal fun isInternetAvailable(connectivityManager: ConnectivityManager): Boolean {
     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
         // The getActiveNetwork() API was introduced in API version 23.
-        val network = connectivityManager.activeNetwork
+        val network = connectivityManager.activeNetwork ?: return false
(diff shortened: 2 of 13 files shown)


code/core/src/test/java/com/adobe/marketing/mobile/services/NetworkServiceTests.java
```diff
@@ -39,12 +46,60 @@ public class NetworkServiceTests {
...
     @Before
     public void setup() throws Exception {
         networkService = new NetworkService(currentThreadExecutorService());
         ServiceProvider.getInstance().setDeviceInfoService(deviceInfoService);
+        ServiceProvider.getInstance().setAppContextService(appContextService);
+        when(appContextService.getConnectivityManager()).thenReturn(null);
+    }
+
+    @Test(timeout = 1000L)
+    public void testConnectAsync_InternetNotAvailable() throws InterruptedException {
```
prudrabhat, line 64: IMO these _functional_ tests add very limited value given that the `AppContextService` and `NetworkUtils` are being mocked any way. So the only functional feature usage from the test suite is waiting for the network request related log to be made.

I recommend adding unit tests for `NetwtorkService` for a similar verification instead by ensuring that the executor service does spin up a thread if there is no network. This can be done by mocking the ExecutorService and the NetworkUtils.kt directly. That way we would not need to add the additional test dependencies.

yangyansong-adbe: 1. `testConnectAsync_InternetNotAvailable` mainly tests if the network callback is invoked with the null connection (`assertNull(connection);`) when the network is unavailable. It's the main contract between the Network Services and its client. This is the basic test case that we should add for sure. 
2. I agree that we can add some logic to check `ExecutorService` execution in these tests.

code/core/src/phone/java/com/adobe/marketing/mobile/internal/util/NetworkUtils.kt
```diff
@@ -0,0 +1,34 @@
...
+import android.net.NetworkCapabilities
+import android.os.Build
+
+/**
+ * Checks if the network is configured to reach the general Internet.
+ *
+ * @param connectivityManager the [ConnectivityManager] to use to check the network status
+ * @return `true` if the network is configured to reach the general Internet, `false` otherwise
+ */
+fun isInternetAvailable(connectivityManager: ConnectivityManager): Boolean {
```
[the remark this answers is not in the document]
yangyansong-adbe: I'd prefer to keep it as an internal utility because it could be used in other places. Also, it's really hard to test a private static method in test code.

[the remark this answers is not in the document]
prudrabhat: Got it. In that case can we make this `internal` and add `@JvmName("isInternetAvailable")` to the method and 
`@file:JvmName("NetworkUtils")` so that you can directly use this in Java internally  (without the *Kt suffix)

code/core/src/test/java/com/adobe/marketing/mobile/services/NetworkServiceTests.java
prudrabhat: For usecase 1 We can still test that the callback is invoked with `null` if we mock the network request and response below:
```
networkService.connectAsync(mockRequest, mockResponseCallback)
verify(mockResponseCallback, times).call(null)
```

For verifying that the network request is made when internet is available it is sufficient to check if the executor service submits a thread.

code/core/src/phone/java/com/adobe/marketing/mobile/internal/util/NetworkUtils.kt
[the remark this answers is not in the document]
yangyansong-adbe: Will do.

code/core/src/test/java/com/adobe/marketing/mobile/services/NetworkServiceTests.java
yangyansong-adbe: I think it's no harm to verify both. I will update the PR to include those changes.  Let me know if you have other thoughts?

code/core/src/phone/java/com/adobe/marketing/mobile/internal/util/NetworkUtils.kt
```diff
@@ -0,0 +1,36 @@
...
+ * Checks if the network is configured to reach the general Internet.
+ *
+ * @param connectivityManager the [ConnectivityManager] to use to check the network status
+ * @return `true` if the network is configured to reach the general Internet, `false` otherwise
+ */
+@JvmName("isInternetAvailable")
+internal fun isInternetAvailable(connectivityManager: ConnectivityManager): Boolean {
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.M) {
+        // The getActiveNetwork() API was introduced in API version 23.
+        val network = connectivityManager.activeNetwork
```
praveek, line 29: nit: val network = connectivityManager.activeNetwork ?: return false

code/core/src/phone/java/com/adobe/marketing/mobile/services/NetworkService.java
```diff
@@ -54,6 +56,20 @@ class NetworkService implements Networking {
...
+            if (!NetworkUtils.isInternetAvailable(connectivityManager)) {
+                Log.trace(ServiceConstants.LOG_TAG, TAG, "The Android network is down.");
+                callback.call(null);
+                return;
+            }
+        } else {
+            Log.warning(
+                    ServiceConstants.LOG_TAG,
+                    TAG,
+                    "ConnectivityManager is null, it's not able to check the network condition.");
```
praveek, line 71: Can you make it a debug message?
nit: "ConnectivityManager instance is null. Unable to the check the network condition."

code/testapp/src/main/java/com/adobe/marketing/mobile/core/testapp/services/NetworkServiceCard.kt
```diff
@@ -0,0 +1,52 @@
...
+            verticalArrangement = Arrangement.Center,
+            horizontalAlignment = Alignment.CenterHorizontally
+        ) {
+            Text(text = "Network Services", modifier = Modifier.padding(16.dp))
+
+            Button(onClick = {
+                val request = NetworkRequest("https://www.adobe.com", HttpMethod.GET, null, null, 5000,5000)
+                ServiceProvider.getInstance().networkService?.connectAsync(request){ connection ->
+                    val status = if (connection != null) "valid connection" else "null connection"
+                    showAlert("Privacy Status: $status")
```
praveek, line 44: Connection status.
nit: I also suggest also showing the http status code if the connection is not null.

code/core/build.gradle.kts
```diff
@@ -43,3 +43,8 @@ apiValidation {
             "com.adobe.marketing.mobile.core.BuildConfig"
     ))
 }
+
+dependencies {
+    androidTestImplementation("org.mockito:mockito-core:5.3.1")
```
praveek, line 48: Can you use the dependency from [BuildConstants](https://github.com/adobe/aepsdk-commons/blob/dev/android/aepsdk-gradle-plugin/aepsdk-gradle-plugin/src/main/kotlin/com/adobe/marketing/mobile/gradle/BuildConstants.kt#L75) until we automatically add this from our plugin?

yangyansong-adbe: It's not included in the AEP plugin's instrumentation tests. I will add a TODO that reminds us of adding it to the plugin later.

praveek (approved):

prudrabhat (approved):

PR #617, opened 2024-02-29
[Android 3.0.0] Adopt the network service changes in Identity

yangyansong-adbe:
The Identity extension is updated to adopt the network service changes added in Android Core 3.0.0. (https://github.com/adobe/aepsdk-core-android/pull/612) 

The test cases mentioned below have been tested on both the Android device (API level 34) and the emulator (API level 21/34).
1. When the device’s network is **online**, call `setPushIdentifier` API multiple times and check the requests are sent out correctly. (monitored through Android Studio’s network inspector view)
2. When the device’s network is **offline**, call `setPushIdentifier` API multiple times (requests should be queued), and there is no network connection established (monitored through Android Studio’s network inspector view)
3. The queued requests are sent out correctly after the device’s network is back **online**.
4. New requests can be sent out immediately if the device’s network is still **online**.

Simulated device/emulator online and offline situations by using:
- Turn on/off the airplane mode
- Turn on/off Wifi+Cellular

### DIFF
--- a/code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
+++ b/code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
@@ -912,6 +912,7 @@ public final class IdentityExtension extends Extension {
             final String urlString =
                     buildURLString(
                             currentCustomerIds, dpids, latestValidConfig, didAdidConsentChange);
+            // URLBuilder will return null, if the final URL is invalid. Drop the hit in that case.
             if (urlString != null) {
                 IdentityHit hit = new IdentityHit(urlString, event);
                 hitQueue.queue(hit.toDataEntity());
--- a/code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
+++ b/code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
@@ -628,6 +628,11 @@ public final class IdentityExtension extends Extension {
                 networkRequest,
                 connection -> {
                     if (connection == null) {
+                        Log.trace(
+                                IdentityConstants.LOG_TAG,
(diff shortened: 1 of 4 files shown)


code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
```diff
@@ -907,8 +912,17 @@ boolean handleSyncIdentifiers(final Event event, final boolean forceSync) {
             final String urlString =
                     buildURLString(
                             currentCustomerIds, dpids, latestValidConfig, didAdidConsentChange);
-            IdentityHit hit = new IdentityHit(urlString, event);
-            hitQueue.queue(hit.toDataEntity());
+            // URLBuilder will return null, if the final URL is invalid. Drop the hit in that case.
+            if (urlString != null) {
```
kevinlind, line 916: Can you add a unit test to verify if the `urlString` is null/malformed then the IdentityHit is not queued?

yangyansong-adbe: I tried, but I can't find an easy way to do that since it's wrapped in a huge method. We can refactor the production code and add tests to cover that use case. At this time, I don't think it is worth spending time on code refactoring. Let me know if you have other thoughts.

kevinlind: Is it possible to set a null/empty `experienceCloud.server` value in Configuration which will trigger a null `urlString`?
https://github.com/yangyansong-adbe/aepsdk-core-android/blob/073a0c633608ed48c512d47f22e24169175eb6ca/code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java#L1964

kevinlind (approved):

PR #618, opened 2024-02-29
Use InAppMessage dimensions relative to host activity

prudrabhat:
## Description
**Problem** : Currently, the size of the InAppMessage is based on the screen height and width. This displays the IAM ass desired for all single activity cases. However, this is not always the case when the activity is not occupying the entire screen.
Such cases will see a full screen message within the activity irrespective of the dimensions of the actual content size of
the activity. So 70% height would be 70% screen height and exceed the activity content view height where as it should have been 70% of the height within the activity.

**Solution**: Use the Use InAppMessage dimensions relative to host activity and not the entire screen.

## Motivation and Context
**Problem** : Currently, the size of the InAppMessage is based on the screen height and width. This displays the IAM ass desired for all single activity cases. However, this is not always the case when the activity is not occupying the entire screen.
Such cases will see a full screen message within the activity irrespective of the dimensions of the actual content size of
the activity. So 70% height would be 70% screen height and exceed the activity content view height where as it should have been 70% of the height within the activity.

**Solution**: Use the Use InAppMessage dimensions relative to host activity and not the entire screen.

## Types of changes

## Checklist:

### DIFF
--- a/code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageContent.kt
+++ b/code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageContent.kt
@@ -16,17 +16,13 @@ import android.webkit.WebView
 import androidx.compose.foundation.gestures.Orientation
 import androidx.compose.foundation.gestures.draggable
 import androidx.compose.foundation.gestures.rememberDraggableState
-import androidx.compose.foundation.layout.height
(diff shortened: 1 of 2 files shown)


prudrabhat:
@spoorthipujariadobe did some initial testing and identified that `LocalConfiguration.current` provides the activity configuration dimensions. So this PR is likely not needed. We can re-open this PR if we find that it is not the case.

PR #624, opened 2024-03-08
Publish Lifecycle, Signal, Identity with `core:3.0.0.beta.1-SNAPSHOT` dependancies

prudrabhat:
## Description
Publish Lifecycle, Signal, Identity use the `coreExtensionVersion` to publish dependencies in POM. 
`addCoreDependency()`  from the aep-library plugin being [used inside the ](https://github.com/adobe/aepsdk-userprofile-android/blob/dev-v3.0.0/code/userprofile/build.gradle.kts#L14) `build.gradle.kts` for extensions do not have a way to know that the current build variant being built. So the SNAPSHOT  suffix is not being added to dependancies
Use "$coreExtensionVersion-SNAPSHPOT" instead.

## Motivation and Context
Publish Lifecycle, Signal, Identity use the `coreExtensionVersion` to publish dependencies in POM. 

`addCoreDependency()`  from the aep-library plugin being [used inside the ](https://github.com/adobe/aepsdk-userprofile-android/blob/dev-v3.0.0/code/userprofile/build.gradle.kts#L14) `build.gradle.kts` for extensions do not have a way to know that the current build variant being built. So the SNAPSHOT  suffix is not being added to dependancies
Use "$coreExtensionVersion-SNAPSHPOT" instead.

## Types of changes

## Checklist:

### DIFF
--- a/code/identity/build.gradle.kts
+++ b/code/identity/build.gradle.kts
@@ -18,7 +18,9 @@ aepLibrary {
         mavenRepoName = identityMavenRepoName
         mavenRepoDescription = identityMavenRepoDescription
         gitRepoName = "aepsdk-core-android"
-        addCoreDependency(coreExtensionVersion)
+
+        // Stop using snapshot version after release
+        addCoreDependency("$coreExtensionVersion-SNAPSHPOT")
     }
 }
 
--- a/code/lifecycle/build.gradle.kts
+++ b/code/lifecycle/build.gradle.kts
@@ -18,7 +18,9 @@ aepLibrary {
         mavenRepoName = lifecycleMavenRepoName
         mavenRepoDescription = lifecycleMavenRepoDescription
         gitRepoName = "aepsdk-core-android"
-        addCoreDependency(coreExtensionVersion)
+
+        // Stop using snapshot version after release
+        addCoreDependency("$coreExtensionVersion-SNAPSHPOT")
     }
 }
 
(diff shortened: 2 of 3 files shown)


code/identity/build.gradle.kts
```diff
@@ -18,7 +18,9 @@ aepLibrary {
         mavenRepoName = identityMavenRepoName
         mavenRepoDescription = identityMavenRepoDescription
         gitRepoName = "aepsdk-core-android"
-        addCoreDependency(coreExtensionVersion)
+
+        // Stop using snapshot version after release
+        addCoreDependency("$coreExtensionVersion-SNAPSHPOT")
```
PravinPK, line 23: if we are planning to hardcode "-SNAPSHPOT".. 
Do you think if it is better to suffix them in the mavenCoreVersion property of gradle.properties file

```
mavenCoreVersion=3.0.0-beta.1-SNAPSHOT
```

prudrabhat: Core project use the same gradle.`coreExtensionVersion` instead of the `mavenCoreVersion` property. This is also used for `core` build.gradle. Adding another `mavenCoreVersion` property is going to be confusing. So,I prefer this approach for this project.
Other extensions can change the `mavenCoreVersion` as needed.

yangyansong-adbe: I agree with @prudrabhat , in most cases, we won't need to use another `mavenCoreVersion` property. Specifying the dependency by adding  `addCoreDependency("$coreExtensionVersion-SNAPSHPOT")` is very straightforward, and we can easily comment it out for future releases.

yangyansong-adbe (approved):

PravinPK (approved):

praveek (approved):

PR #632, opened 2024-03-22
Update getNetworkConnectionStatus implementation to use internal utils

praveek:
Media extension uses this API to check the network connection for offline scenarios. The current implementation uses deprecated APIs and so switch it to use the internal utility for consistency.

### DIFF
--- a/code/core/src/phone/java/com/adobe/marketing/mobile/services/DeviceInfoService.java
+++ b/code/core/src/phone/java/com/adobe/marketing/mobile/services/DeviceInfoService.java
@@ -21,12 +21,12 @@ import android.content.res.AssetManager;
 import android.content.res.Configuration;
 import android.content.res.Resources;
 import android.net.ConnectivityManager;
-import android.net.NetworkInfo;
 import android.os.Build;
 import android.os.Bundle;
 import android.provider.Settings;
 import android.telephony.TelephonyManager;
 import android.util.DisplayMetrics;
+import com.adobe.marketing.mobile.internal.util.NetworkUtils;
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


yangyansong-adbe (approved):

code/core/src/phone/java/com/adobe/marketing/mobile/services/DeviceInfoService.java
```diff
@@ -214,71 +214,15 @@ public String getMobileCarrierName() {
...
+        ConnectivityManager connectivityManager =
+                ServiceProvider.getInstance().getAppContextService().getConnectivityManager();
+        if (connectivityManager == null) {
+            return ConnectionStatus.UNKNOWN;
         }
 
-        return DeviceInfoService.ConnectionStatus.UNKNOWN;
+        return NetworkUtils.isInternetAvailable(connectivityManager)
+                ? ConnectionStatus.CONNECTED
+                : ConnectionStatus.DISCONNECTED;
```
prudrabhat, line 225: I was going to suggest 
```
return (connectivityManager !=null  &&  NetworkUtils.isInternetAvailable(connectivityManager)) ?: ConnectionStatus.CONNECTED : ConnectionStatus.DISCONNECTED
```
but not sure if the existing clause is more clear to read. LGTM either way.

prudrabhat (approved):

code/core/src/phone/java/com/adobe/marketing/mobile/services/DeviceInfoService.java
praveek: `ConnectionStatus.UNKNOWN` makes more sense when connectivityManager is null.

PR #648, opened 2024-04-16
[Refactored] Migrate the out-of-the-box push template handling functionality to the Core extension

rymorale:
Migrates the OOTB push template code to the core extension. The TemplateUtils object will be the interface for extensions to create AEPSDK out-of-the-box push template notifications. Two methods are provided for creating a notification with either a message data map or an intent. The map or intent are then used to create a Notification.Builder object which the calling extension then create a Notification object to be scheduled.

## Types of changes

## Checklist:


code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplate.kt
```diff
@@ -0,0 +1,400 @@
...
+    /**
+     * Constructor to create a push template object from the data payload.
+     *
+     * @param data [Map] of key-value pairs representing the push template data payload
+     */
+    constructor(data: Map<String, String>?) {
+        // fast fail (via IllegalArgumentException) if required data is not present
+        if (data.isNullOrEmpty()) throw IllegalArgumentException("Push template data is null.")
+        this.messageData = data.toMutableMap()
+        val title = DataReader.getString(data, PushTemplateConstants.PushPayloadKeys.TITLE)
```
prudrabhat, line 187: `DataReader.getString()` will throw a `DataReaderException` if the key is not found. This exception does not make much sense to the caller because they do not need to be aware of how the key reading happens in this constructor.  Can you use `DataReader.optString` with a null fallback and then finally throw an `IllegalArgumentException`. 
Same comment for all other cases where `DataReader.get*` is being used.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplate.kt
```diff
@@ -0,0 +1,400 @@
...
+        )
+        isFromIntent = false
+    }
+
+    /**
+     * Constructor to create a push template object from an [Intent].
+     *
+     * @param intent [Intent] containing key value pairs extracted from a push template data payload
+     */
+    constructor(intent: Intent?) {
```
prudrabhat, line 297: Can we have the caller do the null check on the intent and change this constructor to accept a no-null intent.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/BasicTemplateNotificationBuilder.kt
```diff
@@ -0,0 +1,340 @@
...
+
+    private fun getActionButton(jsonObject: JSONObject): AEPPushTemplate.ActionButton? {
+        return try {
+            val label = jsonObject.getString(AEPPushTemplate.ActionButtons.LABEL)
+            if (label.isEmpty()) {
+                Log.debug(PushTemplateConstants.LOG_TAG, SELF_TAG, "Label is empty")
+                return null
+            }
+            var uri: String? = null
+            val type = jsonObject.getString(AEPPushTemplate.ActionButtons.TYPE)
```
spoorthipujariadobe, line 200: Do we need a separate `ActionButtonType` ? Can we use `ActionType.DEEPLINK.toString()` instead?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplate.kt
```diff
@@ -0,0 +1,400 @@
...
+    /** Class representing the action button with label, link and type  */
+    class ActionButton(val label: String, val link: String?, type: String?) {
+        val type: ActionType
+
+        init {
+            this.type = ActionType.valueOf(type ?: ActionType.NONE.name)
+        }
+    }
+
+    internal object ActionButtonType {
```
spoorthipujariadobe, line 42: Can we use `ActionType`.toString() instead?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplate.kt
```diff
@@ -0,0 +1,400 @@
...
+        const val DEEPLINK = "DEEPLINK"
+        const val WEBURL = "WEBURL"
+        const val DISMISS = "DISMISS"
+        const val OPENAPP = "OPENAPP"
+    }
+
+    internal object ActionButtons {
```
spoorthipujariadobe, line 49: How do feel about moving a few of the constants in this class like ActionButtons, NotificationPriority and NotificationVisibility to a constants file?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplateNotificationBuilder.kt
```diff
@@ -0,0 +1,957 @@
...
+        broadcastReceiverName: String?,
+        trackerActivityName: String?,
+        downloadedImageUris: List<String?>,
+        imageCaptions: List<String?>,
+        imageClickActions: List<String?>
+    ): Intent {
+        val clickIntent = Intent(intentAction)
+        broadcastReceiverName?.let {
+            val broadcastReceiver = Class.forName(broadcastReceiverName)
+            clickIntent.setClass(context, broadcastReceiver)
```
prudrabhat, line 263: Can we make the `construct` entry point accept a `Class<out Activity?>` or a `Class<out BroadcastReceiver>` directly, assign it to the intent, and the let the Android system handle their creation.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/FilmstripCarouselTemplateNotificationBuilder.kt
```diff
@@ -0,0 +1,169 @@
...
+        expandedLayout.setOnClickPendingIntent(R.id.rightImageButton, pendingIntentRightButton)
+
+        // create the notification builder with the common settings applied
+        return super.construct(
+            context,
+            pushTemplate,
+            trackerActivityName,
+            smallLayout,
+            expandedLayout
+        )
```
prudrabhat, line 167: After transforming `AEPPushTemplateBuilder` to an object can you apply the default construction properties first and then apply the specific properties. This will prevent any custom properties set in this method from being modified again due to this call.

prudrabhat (commented):
I am about halfway through the PR. Publishing my comments before continuing through the rest.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/BasicTemplateNotificationBuilder.kt
```diff
@@ -0,0 +1,340 @@
...
+            getActionButtonsFromString(actionButtonsString)
+        if (actionButtons.isNullOrEmpty()) {
+            return
+        }
+        for (eachButton in actionButtons) {
+            val pendingIntent: PendingIntent? =
+                if (eachButton.type === AEPPushTemplate.ActionType.DEEPLINK ||
+                    eachButton.type === AEPPushTemplate.ActionType.WEBURL
+                ) {
+                    BasicTemplateNotificationBuilder.createPendingIntent(
```
spoorthipujariadobe, line 141: Do we need to specify `BasicTemplateNotificationBuilder.createPendingIntent`? This method is not overridden in `BasicTemplateNotificationBuilder`, can we simply do `createPendingIntent`?

spoorthipujariadobe: Also can `createPendingIntent` accept `AEPPushTemplate.ActionButton`? If buttons for all notification types would need the same handling, we can move this logic of checking for action uri based on button type to a common place

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplateNotificationBuilder.kt
```diff
@@ -0,0 +1,957 @@
...
+     * @param bodyText `String` containing the body text
+     */
+    internal fun setLargeIcon(
+        notificationBuilder: NotificationCompat.Builder,
+        imageUrl: String?,
+        title: String?,
+        bodyText: String?
+    ) {
+        // Quick bail out if there is no image url
+        if (StringUtils.isNullOrEmpty(imageUrl)) return
```
spoorthipujariadobe, line 575: MobileCore has a method to also `setLargeIconResourceID`. Can we also check that similar to small icon if `imageUrl` is missing?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/ManualCarouselPushTemplate.kt
```diff
@@ -0,0 +1,43 @@
...
+        private set
+    internal var centerImageIndex: Int = PushTemplateConstants.DefaultValues.NO_CENTER_INDEX_SET
+
+    constructor(data: Map<String, String>?) : super(data) {
+        centerImageIndex =
+            if (carouselLayoutType == PushTemplateConstants.DefaultValues.FILMSTRIP_CAROUSEL_MODE) {
+                PushTemplateConstants.DefaultValues.FILMSTRIP_CAROUSEL_CENTER_INDEX
+            } else {
+                PushTemplateConstants.DefaultValues.MANUAL_CAROUSEL_START_INDEX
+            }
```
prudrabhat, line 27: Can you move this into a utility method as it is being used commonly by both the constructors.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateHelpers.kt
```diff
@@ -0,0 +1,313 @@
...
+ * @param cacheService the AEPSDK [CacheService] to use for caching or retrieving
+ * downloaded image assets
+ * @param uri [String] containing an image asset url
+ * @return [Bitmap] containing the image referenced by the `String` uri
+ */
+internal fun downloadImage(cacheService: CacheService, uri: String?): Bitmap? {
+    if (assetCacheLocation.isNullOrEmpty() || uri.isNullOrEmpty()) {
+        return null
+    }
+    val cacheResult = cacheService[assetCacheLocation!!, uri]
```
prudrabhat, line 134: Strong assert `assetCacheLocation!!` is not needed since it is already being done in the previous line.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateHelpers.kt
```diff
@@ -0,0 +1,313 @@
...
+                PushTemplateConstants.PUSH_IMAGE_CACHE
+            )
+    }
+
+private val executor: ExecutorService
+    get() = ExecutorHolder.INSTANCE
+
+private object ExecutorHolder {
+    val INSTANCE: ExecutorService = Executors.newSingleThreadExecutor()
+}
```
prudrabhat, line 73: Can you make this file a utility `object` and then use. 
```
    val  executor: ExecutorService by lazy { Executors.newSingleThreadExecutor()}
```

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateImageHelper.kt
```diff
@@ -0,0 +1,284 @@
...
+package com.adobe.marketing.mobile.services.ui.notification
+
+import android.content.Context
+import android.graphics.Bitmap
+import android.widget.RemoteViews
+import com.adobe.marketing.mobile.core.R
+import com.adobe.marketing.mobile.services.Log
+import com.adobe.marketing.mobile.services.caching.CacheService
+
+internal object PushTemplateImageHelper {
```
prudrabhat, line 21: Now that we have individual objects for each of the builders, can you move the image population to their own builders.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateHelpers.kt
```diff
@@ -0,0 +1,313 @@
...
+private class DownloadImageCallable(val url: String?) :
+    Callable<Bitmap?> {
+    override fun call(): Bitmap? {
+        val bitmap: Bitmap?
+        var connection: HttpURLConnection? = null
+        var inputStream: InputStream? = null
+        try {
+            val imageUrl = URL(url)
+            connection = imageUrl.openConnection() as HttpURLConnection
+            inputStream = connection.inputStream
```
prudrabhat, line 84: A few recommendations
-  use the `use` function to ensure that the connection stream is auto closed. That way we can get rid of the connection close handling in the catch block.
-  use String templates for logging as it is more idiomatic Kotlin
- log the success case only if the bitmap was downloaded successfully.
- disconnect the connection once we are done.

```
        var bitmap: Bitmap? = null
        var connection: HttpURLConnection? = null
        try {
            val imageUrl = URL(url)
            connection = imageUrl.openConnection() as? HttpURLConnection
            connection?.inputStream.use { inputStream ->
                bitmap = BitmapFactory.decodeStream(inputStream)
                Log.trace(
                    PushTemplateConstants.LOG_TAG,
                    SELF_TAG,
                    "Downloaded push notification image from url ($url)"
                )
            }
        } catch (e: IOException) {
            Log.warning(
                PushTemplateConstants.LOG_TAG,
                SELF_TAG,
                "Failed to download push notification image from url ($url). Exception: ${e.message}"
            )
        } finally {
            connection?.disconnect()
        }
        return bitmap
    }
```

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/CarouselPushTemplate.kt
```diff
@@ -0,0 +1,121 @@
...
+
+    // Required, "default" or "filmstrip"
+    internal var carouselLayoutType: String
+        private set
+
+    // Contains the carousel items as a string
+    internal var rawCarouselItems: String
+        private set
+
+    class CarouselItem(
```
prudrabhat, line 40: nit: This can be a `data` class

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/BasicTemplateNotificationBuilder.kt
```diff
@@ -0,0 +1,340 @@
...
+
+        // create the notification builder with the common settings applied
+        val builder =
+            super.construct(context, pushTemplate, trackerActivityName, smallLayout, expandedLayout)
+
+        // add any action buttons defined for the notification
+        addActionButtons(
+            context,
+            trackerActivityName,
+            builder,
```
spoorthipujariadobe, line 86: Instead of passing around the `builder` object, how do you feel about making this an extension function for `NotificationCompat.Builder` inside `AEPPushTemplateNotificationBuilder` ? Also avoids passing around the builder in `LegacyNotificationBuilder`
 Thoughts @prudrabhat?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplateNotificationBuilder.kt
```diff
@@ -0,0 +1,978 @@
...
+import androidx.core.app.NotificationCompat
+import com.adobe.marketing.mobile.MobileCore
+import com.adobe.marketing.mobile.core.R
+import com.adobe.marketing.mobile.services.Log
+import com.adobe.marketing.mobile.services.ServiceProvider
+import com.adobe.marketing.mobile.util.StringUtils
+import com.adobe.marketing.mobile.util.UrlUtils
+import java.util.Random
+
+internal sealed class AEPPushTemplateNotificationBuilder {
```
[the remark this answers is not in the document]
spoorthipujariadobe: Another suggestion. Can we make the methods of this class extension functions for `NotificationCompat.Builder`? Also avoids passing around the builder and we can directly call them for the other template builders?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplate.kt
rymorale: we can definitely move them to the main PushConstants file

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplate.kt
```diff
@@ -0,0 +1,401 @@
...
+    internal var isFromIntent: Boolean?
+
+    /**
+     * Constructor to create a push template object from the data payload.
+     *
+     * @param data [Map] of key-value pairs representing the push template data payload
+     */
+    constructor(data: Map<String, String>?) {
+        // fast fail (via IllegalArgumentException) if required data is not present
+        if (data.isNullOrEmpty()) throw IllegalArgumentException("Push template data is null.")
```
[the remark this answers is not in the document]
rymorale: yes we can make data non null

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplateNotificationBuilder.kt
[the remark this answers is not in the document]
rymorale: the majority of the functions within this object are used to modify the NotificationCompat.Builder object with different settings. adding extension functions within the object allow something like 
```kotlin
builder.setNotificationDeleteAction(
            context,
            trackerActivityClass
        )
```
but the original call was 
```kotlin
setNotificationDeleteAction(
            context,
            trackerActivityClass,
            builder
        )
```
i'll keep an eye out for other areas where i can simplify the code.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplateNotificationBuilder.kt
```diff
@@ -0,0 +1,978 @@
...
+
+        // get custom color from hex string and set it the notification body text
+        if (!expandedBodyTextColor.isNullOrEmpty()) {
+            setElementColor(
+                smallLayout,
+                R.id.notification_body,
+                "#$expandedBodyTextColor",
+                PushTemplateConstants.MethodNames.SET_TEXT_COLOR,
+                PushTemplateConstants.FriendlyViewNames.NOTIFICATION_BODY_TEXT
+            )
```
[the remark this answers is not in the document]
rymorale: double checked the spec and this should be applied to the collapsed and expanded layout. i've updated the comment within AEPPushTemplate to note this as well:
```
    // Optional, Text color for adb_body and adb_body_ex. Represented as six character hex, e.g. 00FF00
```

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplateNotificationBuilder.kt
```diff
@@ -0,0 +1,978 @@
...
+    internal fun createClickIntent(
+        context: Context,
+        pushTemplate: ManualCarouselPushTemplate,
+        intentAction: String,
+        broadcastReceiverName: String?,
+        trackerActivityName: String?,
+        downloadedImageUris: List<String?>,
+        imageCaptions: List<String?>,
+        imageClickActions: List<String?>
+    ): Intent {
```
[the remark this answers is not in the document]
rymorale: this is also used by the`FilmstripCarouselPushTemplateNotificationBuilder`

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/BasicTemplateNotificationBuilder.kt
rymorale: the `BasicTemplateNotificationBuilder` object was added mistakenly. the `createPendingIntent` function had moved classes a couple of times and this is a result of that.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplateNotificationBuilder.kt
rymorale: i can add support in the future for using the large icon for the given resource id. looking at the implementation, its meant to be used for setting the resource id for the notification's primary image. I think this would be useful in displaying a local image in the case where a remote image could not be downloaded.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateHelpers.kt
rymorale: the strong assert is used as assetCacheLocation has an open getter

prudrabhat: I recommend fetching the cache location when required.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplateNotificationBuilder.kt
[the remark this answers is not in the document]
prudrabhat: Then this should not be accepting `pushTemplate: ManualCarouselPushTemplate` as a parameter right?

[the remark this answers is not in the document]
rymorale: the `FilmstripCarouselTemplateNotificationBuilder` and `ManualCarouselPushTemplateNotificationBuilder` both use a `ManualCarouselPushTemplate` as a template. Technically both template types are manual carousels just differing in the way that they display the images.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateHelpers.kt
rymorale: sounds good, will update

prudrabhat (commented):
I am still attempting to identify if we can make the Manual/Filmstrip notification builders more maintainable. Posting comments while I take time to think.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateHelpers.kt
rymorale: updated this to be a utility function named `getAssetCacheLocation` instead

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateHelpers.kt
```diff
@@ -0,0 +1,299 @@
...
+import java.util.concurrent.Callable
+import java.util.concurrent.ExecutorService
+import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
+
+/**
+ * General utility functions to assist in building push template notifications.
+ */
+
+internal object PushTemplateHelpers {
```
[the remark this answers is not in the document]
rymorale: makes sense, will update these

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/TemplateUtils.kt
```diff
@@ -0,0 +1,182 @@
...
+            throw NotificationConstructionFailedException("intent is null, cannot build a notification.")
+        }
+
+        val context = ServiceProvider.getInstance().appContextService.applicationContext
+            ?: throw NotificationConstructionFailedException("Application context is null, cannot build a notification.")
+        val trackerActivityClass =
+            intent.getSerializableExtra(PushTemplateConstants.IntentKeys.TRACKER_NAME) as Class<out Activity>
+        val broadcastReceiverClass =
+            intent.getSerializableExtra(PushTemplateConstants.IntentKeys.BROADCAST_RECEIVER_NAME) as Class<out BroadcastReceiver>
+        val pushTemplateType =
```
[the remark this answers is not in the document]
rymorale: i had missed this but it is definitely feasible thinking about the implementation. i am updating it.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplateNotificationBuilder.kt
```diff
@@ -0,0 +1,989 @@
...
+            // ID.
+            return channelId ?: PushTemplateConstants.DEFAULT_CHANNEL_ID
+        } else {
+            val notificationManager =
+                context.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+            // For Android versions O and above, create a channel if it does not exist and return
+            // the channel ID.
+            channelId?.let { channelIdFromPayload ->
+                // setup a silent channel for notification carousel item change
+                setupSilentNotificationChannel(context, notificationManager, importance)
```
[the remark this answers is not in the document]
rymorale: the silent channel is setup for displaying a notification after it has been interacted with (in the case of a skip left or skip right press for example). we are re-displaying the notification again but with an updated center index and we don't want the re-displayed notification to play a sound when it is shown.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/TemplateUtils.kt
```diff
@@ -0,0 +1,182 @@
...
+import androidx.core.app.NotificationCompat
+import com.adobe.marketing.mobile.services.Log
+import com.adobe.marketing.mobile.services.ServiceProvider
+
+/**
+ * Public facing object to construct a [NotificationCompat.Builder] object for the specified [PushTemplateType].
+ * The [constructNotificationBuilder] methods will build the appropriate notification based on the provided
+ * [AEPPushTemplate] or [Intent].
+ */
+object TemplateUtils {
```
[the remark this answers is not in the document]
rymorale: i like these naming suggestions. I can make these naming updates.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplateNotificationBuilder.kt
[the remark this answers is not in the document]
rymorale: thinking about this further, the silent channel does not need to be setup for every notification. the silent sound is set on a specified notification channel to be used for silent notifications. all silent notifications of the app will use the same notification channel so this setup can be done once only.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplateNotificationBuilder.kt
```diff
@@ -0,0 +1,989 @@
...
+     * received from the payload, Campaign Classic extension's default channel is used. For Android
+     * versions below O, no channel is created. Just return the obtained channel ID.
+     *
+     * @param context the application [Context]
+     * @param channelId `String` containing the notification channel id
+     * @param customSound `String` containing the custom sound to use
+     * @param importance `int` containing the notification importance
+     * @return the channel ID
+     */
+    internal fun createChannelAndGetChannelID(
```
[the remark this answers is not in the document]
rymorale: i've added these functions and remove the silent channel creation from the `createNotificationChannel` function. i have moved the call to `setupSilentNotificationChannel` to the construct function in each builder and now check if the push template was created from an intent before attempting to setup the silent notification channel.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/CarouselTemplateUtil.kt
```diff
@@ -0,0 +1,115 @@
...
+        newIndices.add(newRightIndex)
+        Log.trace(
+            PushTemplateConstants.LOG_TAG,
+            SELF_TAG,
+            "Calculated new indices. New center index is $newCenterIndex, new left index is $newLeftIndex, and new right index is $newRightIndex."
+        )
+        return newIndices
+    }
+
+    internal fun parseCarouselItems(carouselItemsString: String?): MutableList<CarouselPushTemplate.CarouselItem> {
```
prudrabhat, line 73: Is this being used anywhere outside of `CarouselPushTemplate`. Can we make this a private companion method for that class if not.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/CarouselTemplateUtil.kt
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * Calculates a new left, center, and right index given the current center index, total number
+     * of images, and the intent action.
+     *
+     * @param centerIndex [Int] containing the current center image index
+     * @param listSize `Int` containing the total number of images
+     * @param action [String] containing the action found in the broadcast [Intent]
+     * @return [List] containing the new calculated left, center, and right indices
+     */
+    internal fun calculateNewIndices(
```
prudrabhat, line 32: Please see my comment on consolidating `FilmstripCarouselNotificationBuilder` and `ManualCarouselNotificationBuilder` into a single builder `ManualCarouselNotificationBuilder`. If that is done, we can move this method there.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/CarouselTemplateUtil.kt
```diff
@@ -0,0 +1,115 @@
...
+            Log.debug(
+                PushTemplateConstants.LOG_TAG,
+                SELF_TAG,
+                "Failed to parse carousel items from the push template: ${e.localizedMessage}"
+            )
+        }
+        return carouselItems
+    }
+
+    internal fun getDefaultCarouselIndex(carouselLayoutType: String): Int {
```
prudrabhat, line 108: Since this is only used for `ManualCarouselTemplate` can you make it a utility in its companion object.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/builders/AepPushNotificationBuilder.kt
```diff
@@ -0,0 +1,971 @@
...
+            actionID
+        )
+
+        // adding tracking details
+        return TaskStackBuilder.create(context)
+            .addNextIntentWithParentStack(intent)
+            .getPendingIntent(
+                Random().nextInt(),
+                PendingIntent.FLAG_UPDATE_CURRENT or PendingIntent.FLAG_IMMUTABLE
+            )
```
prudrabhat, line 547: TaskStackBuilder usage resulted in a bug recently in Messaging which was fixed. Can you apply the same fix as here please - https://github.com/adobe/aepsdk-messaging-android/pull/242/files

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/builders/FilmstripCarouselNotificationBuilder.kt
```diff
@@ -0,0 +1,288 @@
...
+     *
+     * @param cacheService the [CacheService] used to cache the downloaded images
+     * @param items the list of [CarouselPushTemplate.CarouselItem] objects to be displayed in the filmstrip carousel
+     * @return a [Map] containing the downloaded images, image URIs, image captions, and image click actions for each carousel item
+     */
+    private fun downloadFilmstripImages(
+        cacheService: CacheService,
+        items: List<CarouselPushTemplate.CarouselItem>
+    ): Map<String, List<Any?>> {
+        val downloadedImages = mutableListOf<Bitmap?>()
```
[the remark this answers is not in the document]
rymorale: yes, we can retrieve the bitmaps via the uri's. i'll update this.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/builders/FilmstripCarouselNotificationBuilder.kt
```diff
@@ -0,0 +1,288 @@
...
+import com.adobe.marketing.mobile.services.ui.notification.NotificationConstructionFailedException
+import com.adobe.marketing.mobile.services.ui.notification.PushTemplateConstants
+import com.adobe.marketing.mobile.services.ui.notification.PushTemplateImageUtil
+import com.adobe.marketing.mobile.services.ui.notification.models.CarouselPushTemplate
+import com.adobe.marketing.mobile.services.ui.notification.models.ManualCarouselPushTemplate
+
+/**
+ * Object responsible for constructing a [NotificationCompat.Builder] object containing a manual filmstrip carousel template notification.
+ */
+internal object FilmstripCarouselNotificationBuilder {
```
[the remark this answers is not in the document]
rymorale: makes sense, will update

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/builders/FilmstripCarouselNotificationBuilder.kt
```diff
@@ -0,0 +1,288 @@
...
+            val notificationManager =
+                context.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+            AepPushNotificationBuilder.setupSilentNotificationChannel(
+                notificationManager,
+                pushTemplate.getNotificationImportance()
+            )
+        }
+
+        // create the notification channel if needed
+        val channelIdToUse = AepPushNotificationBuilder.createChannel(
```
spoorthipujariadobe, line 80: Should this be inside `else` block so as to not replace the silent channel created on line 73?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/builders/AepPushNotificationBuilder.kt
```diff
@@ -0,0 +1,971 @@
...
+
+    /**
+     * Creates a silent notification channel if the device is running on Android O or higher. If the
+     * channel already exists, the same channel is used.
+     *
+     * @param notificationManager the [NotificationManager]
+     * @param importance [Int] containing the notification importance
+     */
+    @RequiresApi(Build.VERSION_CODES.O)
+    internal fun setupSilentNotificationChannel(
```
spoorthipujariadobe, line 926: Can we achieve the same with `createNotificationChannel` method in this class? We can modify `createNotificationChannel` to also accept channel name and call like so

```
createNotificationChannel(
        context,
        PushTemplateConstants.DefaultValues.SILENT_NOTIFICATION_CHANNEL_ID,
        PushTemplateConstants.SILENT_CHANNEL_NAME,
        PushTemplateConstants.DefaultValues.SILENT_SOUND,
        importance
)
```

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/builders/AepPushNotificationBuilder.kt
```diff
@@ -0,0 +1,971 @@
...
+            importance
+        )
+        // return the id of the channel.
+        return channelIdToUse
+    }
+
+    @RequiresApi(Build.VERSION_CODES.O)
+    private fun createNotificationChannel(
+        context: Context,
+        notificationManager: NotificationManager,
```
spoorthipujariadobe, line 884: Don't think we need to accept this as a parameter. Since we already have the context we can do 
`val notificationManager = context.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager`

prudrabhat: If a previous call to this method already has/needs the system service always, I prefer having this as a parameter rather than re-fetching.

spoorthipujariadobe: It is currently a private method called only by `createChannel` which always refetches it. We can move it out to the builder class if it makes more sense?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/CarouselTemplateUtil.kt
prudrabhat: I am wondering if we can split this into two methods `calculateNextIndices` and `calculatePreviousIndices`. Since we are concerned with only 3 images, we can also use the `Triple` structure in Kotlin.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/builders/FilmstripCarouselNotificationBuilder.kt
rymorale: the silent channel creation has been moved out of the call to `createChannel`

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/builders/AepPushNotificationBuilder.kt
rymorale: yes, a common channel creation method can be used if given the parameters to use. i'll update.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/builders/ManualCarouselNotificationBuilder.kt
```diff
@@ -0,0 +1,442 @@
...
+import com.adobe.marketing.mobile.services.ui.notification.models.ManualCarouselPushTemplate
+
+/**
+ * Object responsible for constructing a [NotificationCompat.Builder] object containing a manual or filmstrip carousel push template notification.
+ */
+internal object ManualCarouselNotificationBuilder {
+    private const val SELF_TAG = "ManualCarouselTemplateNotificationBuilder"
+
+    @Throws(NotificationConstructionFailedException::class)
+    fun construct(
```
prudrabhat, line 41: Amazing that we have manual and filmstrip consolidated!! I feel that `construct` now has a lot of clauses and it may be better to split them in to single responsibility methods just as we do for other builders. That way each of the methods can have their own logic while maintaining common structure.



```
fun construct(
    context: Context,
    pushTemplate: ManualCarouselPushTemplate,
    trackerActivityClass: Class<out Activity>?,
    broadcastReceiverClass: Class<out BroadcastReceiver>?
): NotificationCompat.Builder {

    val channelIdToUse = setupNotificationChannel(context, pushTemplate)
    val smallLayout = RemoteViews(context.packageName, R.layout.push_template_collapsed)
    val expandedLayout = getExpandedLayout(context, pushTemplate)

    val notificationBuilder = AepPushNotificationBuilder.construct(....)

    val downloadedCarouselItems = downloadCarouselItems()
    setupCarouslelImages(....)
    setupNavigationButtons(...)
    
    return notificationBuilder;
}
```

```
private fun setupCarouselImages(...) {
     if (manual) populateManualCarousel()  
    else populateFilmStrip()
}
```

```
private fun setupNavigationButtons(.....) {
   //generally don't like using a Pair but seems ok in this specific usage
   Pair clickPair = if (pushTemplate.carouselLayoutType ==PushTemplateConstants.DefaultValues.DEFAULT_MANUAL_CAROUSEL_MODE) {
   Pair(PushTemplateConstants.IntentActions.MANUAL_CAROUSEL_LEFT_CLICKED , 
 PushTemplateConstants.IntentActions.FILMSTRIP_RIGHT_CLICKED)
     } else {
          //....
     }
 
   leftIntent =  // use clickPair.first ...
   rightIntent =  // use clickPair.first ...

}
```


```

File: code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/builders/ManualCarouselNotificationBuilder.kt [hunk 36]
```diff
@@ -0,0 +1,442 @@
...
+            captions,
+            interactionUris
+        )
+        val pendingIntentLeftButton = PendingIntent.getBroadcast(
+            context,
+            0,
+            clickIntent,
+            PendingIntent.FLAG_MUTABLE or PendingIntent.FLAG_UPDATE_CURRENT
+        )
+        clickIntent.setAction(
```
prudrabhat, line 206: Can you use separate intents for left and right clicks.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateConstants.kt
```diff
@@ -0,0 +1,164 @@
...
+        const val DEFAULT_MANUAL_CAROUSEL_MODE = "default"
+        const val FILMSTRIP_CAROUSEL_MODE = "filmstrip"
+        const val CAROUSEL_MINIMUM_IMAGE_COUNT = 3
+        const val MANUAL_CAROUSEL_START_INDEX = 0
+        const val FILMSTRIP_CAROUSEL_CENTER_INDEX = 1
+        const val NO_CENTER_INDEX_SET = -1
+
+        // TODO: revisit this value. should cache time be configurable rather than have a static
+        // value?
+        const val PUSH_NOTIFICATION_IMAGE_CACHE_EXPIRY_IN_MILLISECONDS: Long = 259200000 // 3 days
```
prudrabhat, line 57: - `const val PUSH_NOTIFICATION_IMAGE_CACHE_EXPIRY_IN_MILLISECONDS = TimeUnit.DAYS.toMillis(X)`
- Can we check on what we do on iOS with caching and how long are we caching there?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/builders/AepPushNotificationBuilder.kt
```diff
@@ -0,0 +1,974 @@
...
+import com.adobe.marketing.mobile.services.ServiceProvider
+import com.adobe.marketing.mobile.services.ui.notification.NotificationConstructionFailedException
+import com.adobe.marketing.mobile.services.ui.notification.PushTemplateConstants
+import com.adobe.marketing.mobile.services.ui.notification.PushTemplateImageUtil
+import com.adobe.marketing.mobile.services.ui.notification.models.AepPushTemplate
+import com.adobe.marketing.mobile.services.ui.notification.models.ManualCarouselPushTemplate
+import com.adobe.marketing.mobile.util.UrlUtils
+import java.util.Random
+
+internal object AepPushNotificationBuilder {
```
prudrabhat, line 44: We can do this as a followup task but adding a comment here so that we keep track of it. 
The utilities provided by this builder _assumes_ the id's for various common elements (R.id.basic_small_layout , R.id.notification_title, R.id.notification_body_expanded) are the same across templates. We will need to figure out a way to enforce this somehow either programmatically, structurally in the layout  or via documentation.
Can you add a TODO and a note here for this please.

prudrabhat (commented):
LGTM overall.  👏🏼  I have few comments on method simplification and parameters.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AepNotificationUtil.kt
```diff
@@ -0,0 +1,178 @@
...
+import com.adobe.marketing.mobile.services.ui.notification.models.BasicPushTemplate
+import com.adobe.marketing.mobile.services.ui.notification.models.CarouselPushTemplate
+import com.adobe.marketing.mobile.services.ui.notification.models.ManualCarouselPushTemplate
+
+/**
+ * Public facing object to construct a [NotificationCompat.Builder] object for the specified [PushTemplateType].
+ * The [constructNotificationBuilder] methods will build the appropriate notification based on the provided
+ * [AepPushTemplate] or [Intent].
+ */
+object AepNotificationUtil {
```
praveek, line 35: AEPNotificationUtil

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateConstants.kt
```diff
@@ -0,0 +1,164 @@
...
+    const val SILENT_CHANNEL_NAME = "AEPSDK Silent Push Notifications"
+
+    internal object NotificationAction {
+        const val DISMISSED = "Notification Dismissed"
+        const val OPENED = "Notification Opened"
+        const val BUTTON_CLICKED = "Notification Button Clicked"
+    }
+
+    internal class Tracking private constructor() {
+        internal object Keys {
```
praveek, line 36: How about changing it to internal object TrackingKeys ?

code/core/src/test/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplateTests.kt
```diff
@@ -0,0 +1,365 @@
...
+  of the License at http://www.apache.org/licenses/LICENSE-2.0
+  Unless required by applicable law or agreed to in writing, software distributed under
+  the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR REPRESENTATIONS
+  OF ANY KIND, either express or implied. See the License for the specific language
+  governing permissions and limitations under the License.
+*/
+
+package com.adobe.marketing.mobile.services.ui.notification
+
+import androidx.core.app.NotificationCompat
```
prudrabhat, line 14: I haven't reviewed the test class assuming that there are more to be added. We can add them in a follow up PR if you think this is going to be very large

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AepNotificationUtil.kt
```diff
@@ -0,0 +1,178 @@
...
+    private const val SELF_TAG = "AepNotificationUtil"
+
+    @JvmStatic
+    val notificationCompatPriorityMap: Map<Int, String> = mapOf(
```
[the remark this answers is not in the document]
rymorale: this is used on the caller side like:
```
AepNotificationUtil.getNotificationCompatVisibilityMap().get(notification.getVisibility()));
```
in a situation where the caller needs to convert priority or visibility. these maps can be stored on the caller extension side. removing them.

code/core/src/test/java/com/adobe/marketing/mobile/services/ui/notification/AEPPushTemplateTests.kt
rymorale: yes i will add the tests / update these tests in a follow up PR

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateConstants.kt
rymorale: `const val PUSH_NOTIFICATION_IMAGE_CACHE_EXPIRY_IN_MILLISECONDS = TimeUnit.DAYS.toMillis(X)` is not usable as the compiler complains that the value should be constant:
`Const 'val' initializer should be a constant value`

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AepNotificationUtil.kt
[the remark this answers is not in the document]
spoorthipujariadobe: The constant values to create this mapping on the caller side can't be accessed since `AEPPushTemplate` is an internal seal class and `NotificationPriority` is an internal object.  The constant values are also used internally in `AEPPushTemplate` class. Can we keep this here?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateType.kt
```diff
@@ -0,0 +1,36 @@
...
+  governing permissions and limitations under the License.
+*/
+
+package com.adobe.marketing.mobile.services.ui.notification
+
+/**
+ * Enum class representing the different types of out-of-the-box push templates.
+ */
+internal enum class PushTemplateType(val value: String) {
+    BASIC("basic"), CAROUSEL("car"), INPUT_BOX("input"), UNKNOWN("unknown");
```
praveek, line 18: Are we expecting these values from any service? If not, I would prefer "carousel" instead of "car".

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPNotificationUtil.kt
```diff
@@ -0,0 +1,150 @@
...
+        broadcastReceiverClass: Class<out BroadcastReceiver>?
+    ): NotificationCompat.Builder {
+        val context = ServiceProvider.getInstance().appContextService.applicationContext
+            ?: throw NotificationConstructionFailedException("Application context is null, cannot build a notification.")
+        val pushTemplateType =
+            PushTemplateType.fromString(messageData[PushTemplateConstants.PushPayloadKeys.TEMPLATE_TYPE])
+
+        when (pushTemplateType) {
+            PushTemplateType.BASIC -> {
+                val basicPushTemplate = BasicPushTemplate(messageData as MutableMap<String, String>)
```
praveek, line 52: This is not a safe cast. Why do the templates need a MutableMap? Would it be acceptable to use Map instead?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateImageUtil.kt
```diff
@@ -0,0 +1,240 @@
...
+    /**
+     * Downloads an image using the provided uri `String`. A [Future] task is created to download
+     * the image using a [DownloadImageCallable]. The task is submitted to an [ExecutorService].
+     * @param url [String] containing the image url to download
+     * @return [Bitmap] containing the downloaded image
+     */
+    internal fun download(url: String?): Bitmap? {
+        var bitmap: Bitmap? = null
+        val executorService = executor
+        val downloadTask = executorService.submit(DownloadImageCallable(url))
```
prudrabhat, line 139: Currently download logic waits for each image to download based on the custom network request crafted by this class.  Let us use `NetworkService` to download all the images at once, and wait for a fixed time to see if we have minimum number of images for displaying.

praveek (approved):

praveek (changes requested):
Comments inline

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPNotificationUtil.kt
```diff
@@ -0,0 +1,150 @@
...
+                val carouselOperationMode = carouselPushTemplate.carouselOperationMode
+                val carouselType = carouselPushTemplate.carouselLayoutType
+
+                Log.trace(
+                    PushTemplateConstants.LOG_TAG,
+                    SELF_TAG,
+                    "Building a $carouselType carousel style push notification."
+                )
+
+                return if (carouselOperationMode == PushTemplateConstants.DefaultValues.AUTO_CAROUSEL_MODE) {
```
praveek, line 73: nit: Given the object hierarchy, can you update the creation of carouselPushTemplate and the logic to 

when (carouselPushTemplate) {
    is AutoCarouselPushTemplate -> {
        AutoCarouselNotificationBuilder.construct(
            context,
            carouselPushTemplate,
            trackerActivityClass,
            broadcastReceiverClass
        )
    }
    is ManualCarouselPushTemplate -> {
        ManualCarouselNotificationBuilder.construct(
            context,
            carouselPushTemplate,
            trackerActivityClass,
            broadcastReceiverClass
        )
    }
    else -> 
        NotificationConstructionFailedException
}

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPNotificationUtil.kt
```diff
@@ -0,0 +1,150 @@
...
+import com.adobe.marketing.mobile.services.ui.notification.models.CarouselPushTemplate
+import com.adobe.marketing.mobile.services.ui.notification.models.ManualCarouselPushTemplate
+
+/**
+ * Public facing object to construct a [NotificationCompat.Builder] object for the specified [PushTemplateType].
+ * The [constructNotificationBuilder] methods will build the appropriate notification based on the provided
+ * [AEPPushTemplate] or [Intent].
+ */
+object AEPNotificationUtil {
+    private const val SELF_TAG = "AepNotificationUtil"
```
praveek, line 36: nit: AEPNotificationUtil

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/models/AEPPushTemplate.kt
```diff
@@ -0,0 +1,399 @@
...
+     * Constructor to create a push template object from the data payload.
+     *
+     * @param data [Map] of key-value pairs representing the push template data payload
+     */
+    constructor(data: Map<String, String>?) {
+        // fast fail (via IllegalArgumentException) if required data is not present
+        if (data.isNullOrEmpty()) throw IllegalArgumentException("Push template data is null.")
+        this.messageData = data.toMutableMap()
+        val title = DataReader.optString(data, PushTemplateConstants.PushPayloadKeys.TITLE, null)
+        if (title.isNullOrEmpty()) throw IllegalArgumentException("Required field \"adb_title\" not found.")
```
praveek, line 190: nit: I would prefer re using the constants instead of hardcoded values.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/models/AEPPushTemplate.kt
```diff
@@ -0,0 +1,399 @@
...
+     *
+     * @param intent [Intent] containing key value pairs extracted from a push template data payload
+     */
+    constructor(intent: Intent) {
+        val intentExtras =
+            intent.extras ?: throw IllegalArgumentException("Intent extras are null.")
+        // required values
+        title = intentExtras.getString(PushTemplateConstants.IntentKeys.TITLE_TEXT)
+            ?: throw IllegalArgumentException("Required field \"adb_title\" not found.")
+        body = intentExtras.getString(PushTemplateConstants.IntentKeys.BODY_TEXT)
```
praveek, line 302: Why is there inconsistency between the implementations of the two constructors?
For instance, the other constructor throws an exception when encountering an empty string, while also attempting to read ACC_PAYLOAD_BODY:

```
var bodyText = DataReader.optString(data, PushTemplateConstants.PushPayloadKeys.BODY, null)
        if (bodyText.isNullOrEmpty()) {
            bodyText =
                DataReader.optString(
                    data,
                    PushTemplateConstants.PushPayloadKeys.ACC_PAYLOAD_BODY, null
                )
        }
```

praveek: If this is expected, can you add a comment?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/models/CarouselPushTemplate.kt
```diff
@@ -0,0 +1,158 @@
...
+        // Required, URI to an image to be shown for the carousel item
+        val imageUri: String,
+        // Optional, caption to show when the carousel item is visible
+        val captionText: String?,
+        // Optional, URI to handle when the item is touched by the user. If no uri is provided for the item, adb_uri will be handled instead.
+        val interactionUri: String?
+    )
+
+    constructor(data: Map<String, String>) : super(data) {
+        try {
```
praveek, line 51: nit: I would prefer switching to `optString` and a check to see if the returned value is null or empty.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateImageUtil.kt
```diff
@@ -0,0 +1,240 @@
...
+import java.util.concurrent.TimeUnit
+
+/**
+ * Utility functions to assist in downloading and caching images for push template notifications.
+ */
+
+internal object PushTemplateImageUtil {
+    private const val SELF_TAG = "PushTemplateImageUtil"
+    private const val FULL_BITMAP_QUALITY = 100
+    private const val DOWNLOAD_TIMEOUT = 10
```
praveek, line 43: Update variable names to also include the Unit. `DOWNLOAD_TIMEOUT_SEC`

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/builders/BasicNotificationBuilder.kt
```diff
@@ -0,0 +1,384 @@
...
+        if (pushTemplate.isFromIntent == true && Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
+            val notificationManager =
+                context.getSystemService(Context.NOTIFICATION_SERVICE) as NotificationManager
+            AEPPushNotificationBuilder.setupSilentNotificationChannel(
+                notificationManager,
+                pushTemplate.getNotificationImportance()
+            )
+        }
+
+        // create the notification channel if needed
```
praveek, line 71: Can you remove this comment as channelId is always created?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/builders/AEPPushNotificationBuilder.kt
```diff
@@ -0,0 +1,977 @@
...
+     * already exists, the same channel is used. A default channel ID will be used if no channel ID
+     * is received from the payload.
+     *
+     * @param context the application [Context]
+     * @param channelId [String] containing the notification channel id
+     * @param customSound `String` containing the custom sound to apply on the notification channel
+     * @param importance [Int] containing the notification importance
+     * @return A [String] containing the created or existing channel ID
+     */
+    internal fun createChannel(
```
praveek, line 859: Can you update the function name to something like createChannelIfNeeded / createChannelIfRequired?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/builders/AEPPushNotificationBuilder.kt
```diff
@@ -0,0 +1,977 @@
...
+     * @param context the application [Context]
+     * @param pushTemplate the [ManualCarouselPushTemplate] object containing the manual carousel push template data
+     * @param intentAction [String] containing the intent action
+     * @param broadcastReceiverClass the [Class] of the broadcast receiver to set in the created pending intent
+     * @param downloadedImageUris [List] of String` containing the downloaded image URIs
+     * @param imageCaptions `List` of String` containing the image captions
+     * @param imageClickActions `List` of String` containing the image click actions
+     * @return the created click [Intent]
+     */
+    internal fun createClickIntent(
```
praveek, line 311: Any reason this is defined in `AEPPushNotificationBuilder` and not `ManualCarouselNotificationBuilder`?

praveek: nit: I prefer reorganizing the non-private methods, except construct. They can either be moved to a utility class or defined as extensions for appropriate classes.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/TemplateUtils.kt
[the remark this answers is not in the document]
spoorthipujariadobe: We ended up calling `notification.templates` as `notification.models` @prudrabhat can you confirm if we're all aligned on this?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateType.kt
spoorthipujariadobe: These values are to be sent by the ACC push service. Don't think we can change it at this point

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AEPNotificationUtil.kt
spoorthipujariadobe: Doesn't need to be MutableMap, its only read from. Updating it use Map

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/models/AEPPushTemplate.kt
spoorthipujariadobe: The map data is received from the FCM message. My guess is it can have body text as either the `adb_body` or `_msg`.  But in the latter case, it should be converted in the caller extension to the expected adb format. I will leave this one open to confirm with @rymorale.
The intent is from the pending intent constructed by the notification builder classes so its guaranteed to have the key as `bodyText`

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/builders/AEPPushNotificationBuilder.kt
spoorthipujariadobe: I prefer the extension function approach too, to avoid passing around additional params. I will leave this one open as well to discuss again with @rymorale and @prudrabhat

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/builders/AepPushNotificationBuilder.kt
spoorthipujariadobe: Will address in subsequent PR when moving it to extensions

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/builders/AepPushNotificationBuilder.kt
spoorthipujariadobe: Will address in subsequent PR when moving it to extensions

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AepNotificationUtil.kt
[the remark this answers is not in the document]
spoorthipujariadobe: No longer a problem after moving `NotificationPriority` to `PushTemplateConstants` class

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateImageUtil.kt
spoorthipujariadobe: Will address in a subsequent PR

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AepNotificationUtil.kt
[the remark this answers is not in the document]
spoorthipujariadobe: @prudrabhat can we add this back?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/models/CarouselPushTemplate.kt
```diff
@@ -47,7 +48,8 @@ internal open class CarouselPushTemplate : AEPPushTemplate {
         val interactionUri: String?
     )
 
-    constructor(data: Map<String, String>) : super(data) {
+    @Throws(NotificationConstructionFailedException::class)
```
praveek, line 51: I am not seeing this exception thrown from the constructor.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/models/CarouselPushTemplate.kt
```diff
@@ -112,6 +114,14 @@ internal open class CarouselPushTemplate : AEPPushTemplate {
         private const val SELF_TAG = "CarouselPushTemplate"
         const val MINIMUM_FILMSTRIP_SIZE = 3
 
+        fun createCarouselPushTemplate(data: Map<String, String>): CarouselPushTemplate {
+            val carouselPushTemplate = CarouselPushTemplate(data)
```
praveek, line 118: Instead of constructing the entire object can you just read this value? 
val carouselLayoutType = DataReader.optString(
            data, PushTemplateConstants.PushPayloadKeys.CAROUSEL_LAYOUT, null
 )

praveek:
Can you fix the lint errors?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/models/CarouselPushTemplate.kt
spoorthipujariadobe: Removed

praveek (approved):
Approving this. @spoorthipujariadobe will make the network related changes and other cleanup in a new PR

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/models/AEPPushTemplate.kt
spoorthipujariadobe: Verified with Ryan, the caller extension (ACC) needs to covert all the legacy keys to `adb` prefixed one. Removing the logic

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/TemplateUtils.kt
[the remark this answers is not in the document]
spoorthipujariadobe: Refactored these as `notification.templates`

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/AepNotificationUtil.kt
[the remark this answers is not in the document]
spoorthipujariadobe: Added back as public enums

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateConstants.kt
spoorthipujariadobe: Converted to `val` instead of `const val`

prudrabhat (approved):

PR #651, opened 2024-05-07
Notify PresentationDelegate on all webview uri's

prudrabhat:
## Description
Currently, presentation delegate notifies only the url's opened by the UriService - which is a regression from 2.x (where all successfully opened URI's were notified)
As a solution change the logic to make UiService handle both uri's and url's if the event listener is unable to handle them. Further, notify PresentationDelegate of all uri's opened as part of the interaction.

- Update `handleInAppUri` to `handleInAppUri`
- Remove url validation because we now handle all URI's (handled by UriService)
- Add tests

## Related Issue
(Internal) MOB-20924

## Motivation and Context
Currently, presentation delegate notifies only the url's opened by the UriService which is a regression from 2.x.

Change the logic to make UiService handle both uri's and url's if the event listener is unable to handle them. Further, notify PresentationDelegate of all uri's opened as part of the interaction.

## Types of changes

## Checklist:

### DIFF
--- a/code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/InAppMessagePresentable.kt
+++ b/code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/InAppMessagePresentable.kt
@@ -15,6 +15,7 @@ import android.content.Context
 import android.view.ViewGroup
 import android.webkit.WebSettings
 import android.webkit.WebView
+import androidx.annotation.VisibleForTesting
 import androidx.compose.ui.platform.ComposeView
 import com.adobe.marketing.mobile.services.ui.Alert
 import com.adobe.marketing.mobile.services.ui.InAppMessage
@@ -108,9 +109,9 @@ internal class InAppMessagePresentable(
                         // So always dismiss the message when a gesture is detected
                         dismiss()
 
-                        // If a gesture mapping exists, the notify the listener about the url associated with the gesture
+                        // If a gesture mapping exists, the notify the listener about the uri associated with the gesture
                         inAppMessage.settings.gestureMap[gesture]?.let { link ->
-                            handleInAppUrl(link)
+                            handleInAppUri(link)
                         }
                     }
                 )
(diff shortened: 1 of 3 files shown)


code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/InAppMessagePresentable.kt
```diff
@@ -108,9 +109,9 @@ internal class InAppMessagePresentable(
                         // So always dismiss the message when a gesture is detected
                         dismiss()
 
-                        // If a gesture mapping exists, the notify the listener about the url associated with the gesture
+                        // If a gesture mapping exists, the notify the listener about the uri associated with the gesture
                         inAppMessage.settings.gestureMap[gesture]?.let { link ->
-                            handleInAppUrl(link)
+                            handleInAppUri(link)
```
prudrabhat, line 114: We do have instrumentation tests that cover part of this logic partially. However, we will need to enhance them if possible to identify if the URI is loaded after the interaction.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/InAppMessagePresentable.kt
```diff
@@ -169,32 +170,28 @@ internal class InAppMessagePresentable(
...
-            }
-            true
-        } else {
-            false
+    @VisibleForTesting
+    internal fun handleInAppUri(uri: String): Boolean {
+        // First check if the component that created this message is able to handle the uri.
+        // Otherwise check if this URI can be opened by the utility provider via UriOpening service
+        val handled = inAppMessage.eventListener.onUrlLoading(this@InAppMessagePresentable, uri) ||
+            presentationUtilityProvider.openUri(uri)
```
spoorthipujariadobe, line 187: Does 2.x only notify delegate method if `UriService` is able to open the URI? Wondering if there is a case to also notify the delegate about malformed URI

spoorthipujariadobe (approved):
LGTM, just one small question

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/InAppMessagePresentable.kt
prudrabhat: In 2.x we notify if the [url is loaded successfully](https://github.com/adobe/aepsdk-core-android/blob/main-v2/code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/MessageWebViewClient.java#L162).

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/InAppMessagePresentable.kt
```diff
@@ -169,32 +170,28 @@ internal class InAppMessagePresentable(
         return InAppMessageWebViewClient(
             inAppMessage.settings,
             presentationUtilityProvider
-        ) { url -> handleInAppUrl(url) }
+        ) { url -> handleInAppUri(url) }
```
praveek, line 173: nit: url -> uri as you have renamed in all other places.

praveek (approved):

PR #655, opened 2024-05-08
Split NotificationBuilder methods to extension functions

spoorthipujariadobe:
## Description

- Moved most existing functions (except `construct` and template specific functions) from NotificationBuilder classes to one of these classes
1. Methods related to small icon, large icon (only for legacy notification), sound, notification visibility, CTA button click actions: Extension functions for NotificationCompat.Builder in `NotificationCompatBuilderExtensions.kt`
2. Methods related to text color, background color, large icon, notification click action : Extension functions for RemoteViews in `RemoteViewsExtensions.kt`
3. Methods related to notification channel creation: Extension function for NotificationManager in `NotificationManagerExtensions.kt`
4. Methods for creating all pending intents: `PendingIntentUtils.kt`
5. All other miscellaneous methods used across builders/extension functions: `PushTemplateUtils.kt`
Methods moved around as is in all cases without any logic change, except for `NotificationManagerExtensions.createNotificationChannelIfRequired`. This method consolidates logic from `createChannelIfRequired`, `createNotificationChannel` and `setupSilentNotificationChannel`, previously in `AEPPushNotificationBuilder`

- Moved setting title and body to small and expanded layout into `AEPPushNotificationBuilder`
- Changed where `AEPPushNotificationBuilder.construct` is called in `AutoCarouselNotificationBuilder` and `ManualCarouselNotificationBuilder`. It is now called only after downloading images and only if not falling back to basic template.

## Types of changes

## Checklist:


code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateUtils.kt
```diff
@@ -0,0 +1,150 @@
...
+import android.content.pm.PackageManager
+import android.net.Uri
+import androidx.annotation.VisibleForTesting
+import com.adobe.marketing.mobile.services.Log
+import com.adobe.marketing.mobile.services.ui.notification.templates.AEPPushTemplate
+import org.json.JSONArray
+import org.json.JSONException
+import org.json.JSONObject
+
+internal object PushTemplateUtils {
```
spoorthipujariadobe, line 25: This class has methods that don't necessarily fit in as extension functions but are used across templates. Happy to discuss a more suitable place to put these

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/templates/AEPPushTemplate.kt
```diff
@@ -273,8 +273,6 @@ internal sealed class AEPPushTemplate {
             intentExtras.getString(PushTemplateConstants.IntentKeys.ACTION_TYPE)
                 ?: PushTemplateConstants.ActionType.NONE.name
         )
-        expandedBodyText =
```
spoorthipujariadobe, line 276: Duplicated from L292

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateUtils.kt
prudrabhat: I feel like we
- `getIconWithResourceName` , `getDefaultAppIcon`, `getSoundUriForResourceName` can be grouped as  `AppResourceExtensions` or similar. 
- `isValidIcon`, `getActionButtonsFromString` and `getActionButtonFromJSONObject` as a file private methods in `NotificationCompatBuilderExtension`
let me know what you think,

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/extensions/RemoteViewsExtensions.kt
```diff
@@ -0,0 +1,271 @@
...
+}
+
+/**
+ * Sets the large icon for the provided [RemoteViews] by downloading the image from the provided URL.
+ * If the image cannot be downloaded, the large icon visibility is set to [View.GONE].
+ *
+ * @param largeIcon `String` containing the large icon URL to download and use
+ */
+internal fun RemoteViews.setRemoteLargeIcon(largeIcon: String?) {
+    if (UrlUtils.isValidUrl(largeIcon)) {
```
prudrabhat, line 229: nit: early return here to avoid nested clauses.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/extensions/NotificationManagerExtensions.kt
```diff
@@ -0,0 +1,88 @@
...
+
+    // Create a channel
+    val channel = NotificationChannel(
+        channelIdToUse,
+        if (isFromIntent == true) PushTemplateConstants.SILENT_CHANNEL_NAME else PushTemplateConstants.DEFAULT_CHANNEL_NAME,
+        importance
+    )
+
+    // Add a sound if required.
+    if (isFromIntent == true) {
```
prudrabhat, line 72: nit: sufficient to do `if (isFromIntent)`

prudrabhat:
Can to add tests for the extensions and utils as part of this PR. They seem to be isolated from the broader implementation..

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateUtils.kt
spoorthipujariadobe: Like the idea of `AppResourceExtensions`! But for `getActionButtonsFromString` and `getActionButtonFromJSONObject`, it would be very hard to unit test them as private methods. How about adding it as a method to `AEPPushTemplate.ActionButton` class? `AEPPushTemplate` already have the `actionButtonString` anyway

spoorthipujariadobe: Also thinking it might make more sense to have `ActionButton` as a separate class and maintain a list of action buttons in AEPPushTemplate

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/extensions/NotificationManagerExtensions.kt
spoorthipujariadobe: `isFromIntent` is an optional boolean, so need to make the explicit comparison

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/templates/AEPPushTemplate.kt
```diff
@@ -98,6 +143,10 @@ internal sealed class AEPPushTemplate {
     internal var actionButtonsString: String?
         private set
 
+    // Optional, list of ActionButton for the notification
+    internal var actionButtonsList: List<ActionButton>?
+        private set
+
```
prudrabhat, line 149: Given that this is the base template for all other templates, it should only contain attributes that are common to _all_ templates. I believe not all templates allow action buttons. So I prefer not to have the action button related logic here in this model class.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/extensions/NotificationManagerExtensions.kt
prudrabhat: When can this be null?  The caller should always know whether this is from an intent or not isn't?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PushTemplateUtils.kt
prudrabhat: Given that`AEPPushTemplate ` is the base template for all other templates, it should only contain attributes that are common to all templates. I believe not all templates allow action buttons. So I prefer not to have the action button related logic (except for ActionButton model definition) there.

If testing is the only issue, why not expose them as `internal` and `@VisibleForTesting`? These will anyway be moved to `notification.internal` package when migrating to its own repo.

prudrabhat (changes requested):

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/templates/AEPPushTemplate.kt
spoorthipujariadobe: Action buttons are only supported in basic template type. Moving this to `BasicPushTemplate`

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/extensions/NotificationManagerExtensions.kt
spoorthipujariadobe: The caller (builder) does not know if the data is from intent or map. Only the template class is aware of this. Since we set `isFromIntent` value is both variations of the template constructor, we can make it a non optional value and simplify this comparsion

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/templates/BasicPushTemplate.kt
```diff
@@ -12,10 +12,77 @@
...
 internal class BasicPushTemplate : AEPPushTemplate {
+
+    private val SELF_TAG = "BasicPushTemplate"
+
+    /** Class representing the action button with label, link and type  */
+    class ActionButton(val label: String, val link: String?, type: String?) {
+        val type: PushTemplateConstants.ActionType
+
+        init {
+            this.type = PushTemplateConstants.ActionType.valueOf(type ?: PushTemplateConstants.ActionType.NONE.name)
```
prudrabhat, line 32: Can you wrap this in a try/catch to swallow `IllegalArgumentException` or `Exception` in case of an invalid `type` string

prudrabhat (approved):

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/notification/PendingIntentUtils.kt
```diff
@@ -0,0 +1,83 @@
...
+
+import android.app.Activity
+import android.app.PendingIntent
+import android.content.Context
+import android.content.Intent
+import java.util.Random
+
+internal object PendingIntentUtils {
+
+    private const val SELF_TAG = "IntentUtils"
```
praveek, line 22: PendingIntentUtils

praveek (approved):

PR #663, opened 2024-05-14
Identity don't dispatch Analytics push event on privacy change

kevinlind:
## Description

Updates the Identity extension to not dispatch the `AnalyticsForIdentityRequest` event when privacy status is changed to optedout.

The Identity extension is expected to dispatch the `AnalyticsForIdentityRequest` event when the push identifier changes or when push tracking is denied the first time. This event tells Analytics to dispatch an ADBINTERNAL:Push event with `a.push.optin` value of either `True` or `False`. The issue is in some cases changing the privacy status to `optedout` would trigger a dispatch of the `AnalyticsForIdentityRequest` event even though `setPushIdentifier` was never called.

This fix changes how Identity handles processing of an privacy status change to `optedout` by clearing the push token and flags instead of calling `updatePushIdentifier` with nil, which would dispatch the event to Analytics.

## Types of changes

## Checklist:

### DIFF
--- a/code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
+++ b/code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
@@ -80,7 +80,6 @@ public final class IdentityExtension extends Extension {
 
     private static final String LOG_SOURCE = "IdentityExtension";
     private HitQueuing hitQueue;
-    private static boolean pushEnabled = false;
     private static final Object pushEnabledMutex = new Object();
     @VisibleForTesting ConfigurationSharedStateIdentity latestValidConfig;
     private final NamedCollection namedCollection;
@@ -1673,7 +1672,7 @@ public final class IdentityExtension extends Extension {
     }
 
     /**
-     * Updates the {@link #pushEnabled} field and dispatches an event to generate a corresponding
+     * Updates the persisted {@code ADOBEMOBILE_PUSH_ENABLED} field and dispatches an event to generate a corresponding
      * Analytics request
      *
      * @param isEnabled whether the user is opted in to receive push notifications
@@ -1726,14 +1725,12 @@ public final class IdentityExtension extends Extension {
                 return false;
             }
 
-            pushEnabled = namedCollection.getBoolean(DataStoreKeys.PUSH_ENABLED, false);
+            return namedCollection.getBoolean(DataStoreKeys.PUSH_ENABLED, false);
         }
-
-        return pushEnabled;
     }
 
     /**
-     * Updates the {@link #pushEnabled} flag in DataStore with the provided value.
+     * Updates the persisted {@code ADOBEMOBILE_PUSH_ENABLED} flag in DataStore with the provided value.
      *
      * @param enabled new push status value to be updated
      */
@@ -1741,6 +1738,11 @@ public final class IdentityExtension extends Extension {
         synchronized (pushEnabledMutex) {
             if (namedCollection != null) {
                 namedCollection.setBoolean(DataStoreKeys.PUSH_ENABLED, enabled);
+                Log.trace(
+                        IdentityConstants.LOG_TAG,
+                        LOG_SOURCE,
+                        "setPushStatus : Push notifications status is now: "
+                                + (enabled ? "Enabled" : "Disabled"));
             } else {
                 Log.trace(
                         IdentityConstants.LOG_TAG,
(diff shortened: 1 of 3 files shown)


code/identity/src/test/java/com/adobe/marketing/mobile/identity/IdentityFunctionalTests.kt
```diff
@@ -2037,4 +2042,178 @@ class IdentityFunctionalTests {
...
+            ).build()
+        )
+
+        countDownLatch.await()
+
+        assertNull(persistedData["ADOBEMOBILE_PUSH_IDENTIFIER"] as? String)
+        assertTrue(persistedData["ADOBEMOBILE_ANALYTICS_PUSH_SYNC"] as? Boolean ?: false)
+        assertFalse(persistedData["ADOBEMOBILE_PUSH_ENABLED"] as? Boolean ?: true)
+    }
+    @Test(timeout = 10000)
```
praveek, line 2145: Can you remove the timeout param? The test does not wait on any asynchronous operation.

code/identity/src/test/java/com/adobe/marketing/mobile/identity/IdentityFunctionalTests.kt
```diff
@@ -2037,4 +2042,178 @@ class IdentityFunctionalTests {
...
+                "com.adobe.eventSource.responseContent"
+            ).setEventData(
+                mapOf("global.privacy" to MobilePrivacyStatus.OPT_OUT.value)
+            ).build()
+        )
+
+        assertNull(identityExtension.mid)
+    }
+
+    @Test(timeout = 10000)
```
praveek, line 2176: Same as above

praveek (approved):

addb (commented):
Changes look good. Let's add unit tests if it is feasible to check resetIdentifier is called and does reset.

code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
```diff
@@ -499,18 +499,8 @@ void handleIdentityRequestReset(@NonNull final Event event) {
...
-        customerIds = null;
-        pushIdentifier = null;
-
-        if (namedCollection != null) {
-            namedCollection.remove(DataStoreKeys.AID_SYNCED_KEY);
-            namedCollection.remove(DataStoreKeys.PUSH_ENABLED);
-        }
 
+        clearIdentifiers();
         savePersistently(); // clear datastore
```
addb, line 504: Should we move savePersistently() to clearIdentifiers() as it is part of syncing cleared values to persistence?

code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
```diff
@@ -1857,17 +1847,7 @@ void processPrivacyChange(final Event event, final Map<String, Object> eventData
...
-            locationHint = null;
-            customerIds = null;
-
-            if (namedCollection != null) {
-                namedCollection.remove(DataStoreKeys.AID_SYNCED_KEY);
-            }
-
-            updatePushIdentifier(null);
+            clearIdentifiers();
             savePersistently(); // clear datastore
```
addb, line 1851: Should we move savePersistently() to clearIdentifiers() as it is part of syncing cleared values to persistence?

addb:
Do we need `pushEnabled` member field L83 if we are using `isPushEnabled()` method when needed?

addb:
> Do we need `pushEnabled` member field L83 if we are using `isPushEnabled()` method when needed?

Do you think we should add `pushEnabled`, `analyticsSynced` as member variable which we load from persistence like other fields like mid and clear similar to other fields on `savePersistently()`. Write now all the keys which are stored on persistence are member variables except for `pushEnabled` and `analyticsSynced`.

addb (approved):

code/identity/src/main/java/com/adobe/marketing/mobile/identity/IdentityExtension.java
addb: @kevinlind What do you think about this change?

kevinlind:
> > Do we need `pushEnabled` member field L83 if we are using `isPushEnabled()` method when needed?
> 
> Do you think we should add `pushEnabled`, `analyticsSynced` as member variable which we load from persistence like other fields like mid and clear similar to other fields on `savePersistently()`. Write now all the keys which are stored on persistence are member variables except for `pushEnabled` and `analyticsSynced`.

This is similar to how we handle the pushEnabled and analyticsSynced flags in iOS, so I'd like to keep it this way for parity.

PR #681, opened 2024-06-17
Detach presentables before re-attaching to top Activity

prudrabhat:
## Description

Currently, the `ComposeView` attached to the `Activity` when presentables are shown is not removed until the activity is destroyed. This is OK for most use-cases that follow the single window mode where the newly resumed activity obscures the previous one, and the presentable redraws on the new activity. 

However, this approach of waiting until the activity is destroyed may result in the presentables being shown on each activity in a multi-window side-by-side mode.
While we do not support multi-window mode, we should ensure that the experience does not deteriorate in such a case. As a solution, change the current detachment logic to to the following:

1.  Track the current activity that the presentable is attached to

2. When presentable is visible on `ActivityA` and another activity `ActivityB` resumes, use the Android activity co-ordination guarantees to detach presentable from Activity-A and re-attach to Activity B.
3. Reset the reference held to the activity that the presentable is attached to when the activity is destroyed.

See More here:
 -  [Configuration Changes](https://developer.android.com/guide/components/activities/state-changes#cco) call `onDestroy` of previous activity before `onResume` of newly created activity
 -  [New activity launches](https://developer.android.com/guide/components/activities/state-changes#covered) always result in `onResume` on new Activity. When the top-most activity is removed, `onResume` on previously paused activity is invoked.    
 -  [Activity co-ordination](https://developer.android.com/guide/components/activities/activity-lifecycle) if `ActivityA` starts `ActivityB`, then `ActivityB.onResume` is invoked before `ActivityA.onPause()`/ `ActivityA.onDestroy()`

## Related Issue
N/A

## How Has This Been Tested?
- Test app
- Scenario test

## Screenshots (if appropriate):
**Before** 
before
**After**
after

## Types of changes

## Checklist:

### DIFF
--- a/code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/common/AEPPresentable.kt
+++ b/code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/common/AEPPresentable.kt
@@ -60,6 +60,14 @@ internal abstract class AEPPresentable<T : Presentation<T>> :
     @VisibleForTesting internal val contentIdentifier: Int = Random().nextInt()
 
     /**
+     * Represents the activity to which the presentable is currently attached.
+     * This SHOULD always be updated whenever the presentable is attached or detached from an
+     * activity. For the sake of maintainability, modify this only in [attach] and [detach] and
+     * limit queries to this field in activity lifecycle methods.
+     */
+    private var attachmentHandle: WeakReference<Activity?> = WeakReference(null)
+
+    /**
      * @param presentation the [Presentation] to be used by this [Presentable]
      * @param presentationUtilityProvider the [PresentationUtilityProvider] to be used to fetch components needed by this [Presentable]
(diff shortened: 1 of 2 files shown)


sbenedicadb (approved):
looks good

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/common/AEPPresentable.kt
```diff
@@ -59,6 +59,14 @@ internal abstract class AEPPresentable<T : Presentation<T>> :
 
     @VisibleForTesting internal val contentIdentifier: Int = Random().nextInt()
 
+    /**
```
prudrabhat, line 62: For some reason Circle CI is not picking up updated coverage  https://app.codecov.io/gh/adobe/aepsdk-core-android/pull/681

praveek (approved):

PR #729, opened 2024-12-17
[MOB-22458][MOB-22611] Cleanup StringEncoder util

praveek:
- Clean up unused functions in StringEncoder, as they were flagged by static analysis tools.
- Extensions like Messaging and Campaign currently use the internal utility. Make StringEncoder a public utility and retain the current class until version 4.x for compatibility.
- Convert some functions to Kotlin and add tests.


code/core/src/main/java/com/adobe/marketing/mobile/util/StringEncoder.kt
```diff
@@ -0,0 +1,56 @@
...
+            }.digest()
+
+            val sha2HexBuilder = StringBuilder()
+            for (aMessageDigest in messageDigest) {
+                val hexString = String.format("%02x", LSB_8_MASK and aMessageDigest.toInt())
+                sha2HexBuilder.append(hexString)
+            }
+            return sha2HexBuilder.toString()
+        } catch (e: NoSuchAlgorithmException) {
+            Log.debug(
```
yangyansong-adbe, line 48: Should we print error logs here?

yangyansong-adbe (approved):

code/core/src/main/java/com/adobe/marketing/mobile/internal/util/StringExtensions.kt
```diff
@@ -0,0 +1,50 @@
...
+ * @return A 32-bit signed integer representing the FNV-1a hash of the input string.
+ */
+private fun computeFnv1aHash(input: String?): Int {
+    val ALL_BITS_ENABLED = 0xFF
+    val PRIME = 0x1000193 // 16777619 as hex
+    val OFFSET = -0x7ee3623b // 2166136261 as hex
+
+    if (input == null) {
+        return 0
+    }
```
prudrabhat, line 30: Will it be simpler if we combine null+empty input validation and move it earlier -   
```
    if (input.isNullOrEmpty()) return 0

    val ALL_BITS_ENABLED = 0xFF
    val PRIME = 0x1000193 // 16777619 as hex
    val OFFSET = -0x7ee3623b // 2166136261 as hex
    var hash = OFFSET
    
    for(...) { ...}
    

```

prudrabhat (approved):

code/core/src/main/java/com/adobe/marketing/mobile/internal/util/StringExtensions.kt
praveek: We don't return zero if the string is empty. The hashes are different based on length.

praveek: I also realized that we should have simply kept var hash = OFFSET, as our implementation diverges from the original algorithm when generating hashes for strings containing only empty characters. However, since this is a corner case, I have decided not to modify the implementation and will leave it as it currently exists in our SDK.

PR #732, opened 2025-01-29
Simplified initialization for AEP SDK

praveek:
Add simplified initialization for AEP SDK.

- Added MobileCore.initialize(..) API and overloads
- Added InitOptions class 
- The new API automatically registers bundled extensions, sets up lifecycle tracking, and configures the SDK.
- Refactored related classes and tests.


code/core/src/main/java/com/adobe/marketing/mobile/ExtensionDiscovery.kt
```diff
@@ -0,0 +1,60 @@
+/*
+  Copyright 2024 Adobe. All rights reserved.
```
yangyansong-adbe, line 2: Let's update the newly created license header to 2025.

code/core/src/main/java/com/adobe/marketing/mobile/ExtensionDiscovery.kt
```diff
@@ -0,0 +1,60 @@
...
+import android.content.Context
+import android.content.pm.PackageManager
+import com.adobe.marketing.mobile.internal.CoreConstants
+import com.adobe.marketing.mobile.services.Log
+
+/**
+ * Responsible for discovering and retrieving a list of [Extension] classes
+ * that are bundled with the app.
+ */
+internal class ExtensionDiscovery {
```
yangyansong-adbe, line 24: Do we plan to add more functions to this class? If not, only exposing an internal function is a good choice.

code/core/src/main/java/com/adobe/marketing/mobile/ExtensionDiscovery.kt
```diff
@@ -0,0 +1,60 @@
...
+    private val SERVICE_NAME = "com.adobe.marketing.mobile.ExtensionDiscoveryService"
+    private val LOG_TAG = "ExtensionDiscovery"
+
+    fun getExtensions(context: Context): List<Class<out Extension>> {
```
yangyansong-adbe, line 28: Reflection-based scanning can be resource-intensive; consider caching results in case this function is called multiple times by accident.

code/core/src/phone/java/com/adobe/marketing/mobile/MobileCoreInitializer.kt
```diff
@@ -0,0 +1,215 @@
...
+    }
+}
+
+/**
+ * Responsible for tracking the application lifecycle and automatically invoking lifecycle APIs.
+ * The lifecycle extension has internal logic to ignore lifecycle calls made during activity switching.
+ * This class handles activity counting for multi-resume scenarios and relies on the lifecycle extension
+ * to drop repeated calls as necessary.
+ */
+internal class LifecycleTracker(val additionalContextData: Map<String, String>?) : App.ActivityLifecycleCallbacks {
```
prudrabhat, line 190: Can we add a small note for reference that the callbacks here are invoked on the main thread and hence why they are serial.

prudrabhat (approved):

kevinlind (approved):
Looks great! `InitOptions.lifecycleAutomaticTrackingEnabled` is named differently than in the spec, but it follows the convention for booleans. I'll update the iOS implementation to match.

code/core/src/main/java/com/adobe/marketing/mobile/ExtensionDiscovery.kt
praveek: Currently, we don't call it more than one. I'll add caching if that situation arises.

PR #735, opened 2025-02-07
Update automatic lifecycle tracking during simplified initialization to use ProcessLifecycleOwner

praveek:
Tracking lifecycle using ActivityLifecycleCallbacks requires the API to be called before the first onResume. This approach offers flexibility since ProcessLifecycleOwner always delivers the first lifecycle onResume when attaching the observer, if the app is in the foreground.


prudrabhat (approved):

code/core/src/phone/java/com/adobe/marketing/mobile/MobileCoreInitializer.kt
```diff
@@ -196,20 +218,15 @@ internal class MobileCoreInitializer(
  * to drop repeated calls as necessary.
  * The callback is invoked synchronously from the registered [Application.ActivityLifecycleCallbacks].
```
prudrabhat, line 219: nit: Needs doc update.

PR #748, opened 2025-03-18
[MOB-22892] Code changes computing webView in accordance with content

siddique-adobe:
## Description

Adding capability for setting the height of the webView frame from custom HTML

### Screenshots: 
Screenshot_20250422_154258

Screenshot 2025-04-22 at 4 56 24 PM

Screenshot 2025-04-22 at 4 56 47 PM

Screenshot_20250422_165139 at 5 00 47 PM

## Related Issue

https://jira.corp.adobe.com/browse/MOB-22892
https://jira.corp.adobe.com/browse/MOB-23078

## Types of changes

## Checklist:


prudrabhat:
@siddique-adobe - I am running late on this review. Will take a look at this PR by  May 24.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageFrame.kt
```diff
@@ -84,9 +92,28 @@ internal fun MessageFrame(
     // multi-window modes.
     val density = LocalDensity.current
     val contentView = currentActivity.findViewById<View>(android.R.id.content)
+    var contentHeightFromJs by remember { mutableStateOf(0) }
+    if (inAppMessageSettings?.fitToContent == true) {
+        LaunchedEffect(Unit) {
+            suspendCoroutine<Int> { continuation ->
+                inAppMessageEventHandler.evaluateJavascript("getContentHeight();") { result ->
```
prudrabhat, line 99: This workflow assume that the `getContentHeight()` java script function will always exist in the HTML being displayed when this setting is enabled. 
- Does the authoring UI automatically include this as part of the in-app message when fitToContent is checked? 
- How are we dealing with custom HTML's? i.e How does the implementer know to include this method?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageFrame.kt
```diff
@@ -84,9 +92,28 @@ internal fun MessageFrame(
     // multi-window modes.
     val density = LocalDensity.current
     val contentView = currentActivity.findViewById<View>(android.R.id.content)
+    var contentHeightFromJs by remember { mutableStateOf(0) }
+    if (inAppMessageSettings?.fitToContent == true) {
+        LaunchedEffect(Unit) {
+            suspendCoroutine<Int> { continuation ->
```
prudrabhat, line 98: From what I see, this coroutine may run before the `MessageScreen` which hosts the WebView is executed. Is my understanding correct? Is it possible to evaluate javascript on a web-view even before it is loaded? If not, we should ensure that this coroutine will run after the webview loads.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageFrame.kt
```diff
@@ -117,7 +144,7 @@ internal fun MessageFrame(
             modifier = Modifier
                 .fillMaxSize()
                 .onPlaced {
-                    heightDp.value = with(density) { ((contentView.height.toDp() * inAppMessageSettings.height) / 100) }
+                    if (inAppMessageSettings?.fitToContent == false) heightDp.value = with(density) { ((contentView.height.toDp() * inAppMessageSettings.height) / 100) }
```
prudrabhat, line 147: This logic ensures that any orientation/configuration change leading to recomposition of the in-app message results in a re-calculation of dimensions. Is this not required for fit to content? What would be the downside of having this run always?

prudrabhat (commented):
Can you add the following details in you PR to help better understand the impact of this change :
- How is the `fitToContent` setting set and controlled? What component is responsible for adding the related javascript in HTML.
- Screenshots of following in-app behaviour before and after the change
      - non-scrollable content with full screen IAM and Modal IAM
      - scrollable content with full screen IAM and Modal IAM
      - IAM when offsets are applied
      - All above cases when device is rotated.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageFrame.kt
siddique-adobe: 1. Yes
2. The continuation.resume() will be called after the JavaScript evaluation completes

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageFrame.kt
siddique-adobe: The `continuation.resume()` will be called after the JavaScript evaluation completes

spoorthipujariadobe (commented):
I share some of the concerns as @prudrabhat around how the `getContentHeight` javascript function would get added to custom HTMLs.
Also I don't see any tests added for this functionality. Do you plan to add them? Could you also share if you have tested this manually for:
* App in split screen mode
* App with edge-to-edge enabled
* IAM with UI takeover enabled (We recently discovered an [issue](https://jira.corp.adobe.com/browse/MOB-23069) around the Dialog component used when UI takeover is enabled)

code/core/src/main/java/com/adobe/marketing/mobile/services/ui/message/InAppMessageSettings.kt
```diff
@@ -48,7 +48,8 @@ class InAppMessageSettings private constructor(
     val cornerRadius: Float,
     val shouldTakeOverUi: Boolean,
     val assetMap: Map<String, String>,
-    val gestureMap: Map<MessageGesture, String>
+    val gestureMap: Map<MessageGesture, String>,
+    val fitToContent: Boolean? = null
```
prudrabhat, line 52: Given that this is a Boolean and the backward compatible or absent configurations can have `fitToContent` as false, does it make sense to make this required and default to false in `InAppMessageSettings`?

```suggestion
`val fitToContent: Boolean = false`
```

nit: Also, I am not quite sure if `fitToContent` is the right name here. It does not give enough information on what content it is fitting to. Can we think of a few alternatives?

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageContent.kt
```diff
@@ -55,6 +58,18 @@ internal fun MessageContent(
                     ViewGroup.LayoutParams.MATCH_PARENT
                 )
 
+                inAppMessageEventHandler.handleJavascriptMessage("inAppContentHeightHandler") { message ->
+                    try {
+                        val height = message.toIntOrNull() ?: 0
```
prudrabhat, line 63: Careful here! If the transformation to Int fails and we default to 0, and based on the current flow this value is directly used for calculation - you will risk not showing the WebView at all.

I recommend that we use a negative constant a contract and ensure we are verifying that it was a failure in `MessageFrame `.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageFrame.kt
```diff
@@ -174,7 +179,17 @@ internal fun MessageFrame(
...
             ) {
-                MessageContent(Modifier.height(heightDp.value).width(widthDp.value), inAppMessageSettings, onCreated)
+                MessageContent(
+                    Modifier
+                        .height(heightDp.value)
+                        .width(widthDp.value),
+                    inAppMessageSettings,
+                    inAppMessageEventHandler,
+                    onHeightReceived = { height ->
+                        if (inAppMessageSettings.fitToContent == true) heightDp.value = height.dp
```
prudrabhat, line 189: Is this `height` received from the Javascript already adjusted for density and recalculated? I assume not, because the js layer will likely not have those details...

prudrabhat (changes requested):

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageFrame.kt
siddique-adobe: removed @prudrabhat

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageFrame.kt
siddique-adobe: removed @prudrabhat

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageContent.kt
siddique-adobe: removed the logic from here and handled gracefully in `MessageFrame.kt`

code/core/src/main/java/com/adobe/marketing/mobile/services/ui/message/InAppMessageSettings.kt
siddique-adobe: done @prudrabhat

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageContent.kt
```diff
@@ -55,6 +58,18 @@ internal fun MessageContent(
                     ViewGroup.LayoutParams.MATCH_PARENT
                 )
 
+                inAppMessageEventHandler.handleJavascriptMessage("inAppContentHeightHandler") { message ->
```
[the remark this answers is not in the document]
siddique-adobe: Got your point, initially i thought of doing that but since the setup was already present in `DefaultInAppMessageEventHandler.kt`, that why i thought of reusing the same,
Now i've as per your comment please check @prudrabhat

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageFrame.kt
```diff
@@ -117,8 +117,12 @@ internal fun MessageFrame(
             modifier = Modifier
                 .fillMaxSize()
                 .onPlaced {
-                    heightDp.value = with(density) { ((contentView.height.toDp() * inAppMessageSettings.height) / 100) }
-                    widthDp.value = with(density) { ((contentView.width.toDp() * inAppMessageSettings.width) / 100) }
+                    if (!inAppMessageSettings.fitToContent) {
+                        heightDp.value =
+                            with(density) { ((contentView.height.toDp() * inAppMessageSettings.height) / 100) }
```
prudrabhat, line 122: This will regress the existing behavior of recalculating the height on placement when `fitToContent` is disabled, doesn't it? Can we ensure that we set **both** the height and width to the desired value irrespective of the fit to content feature.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageFrame.kt
```diff
@@ -174,7 +178,47 @@ internal fun MessageFrame(
...
+                                    ServiceConstants.LOG_TAG,
+                                    "MessageFrame",
+                                    "Invalid height value received: $height"
+                                )
+
+                                height.toInt() <= 0 -> Log.warning(
+                                    ServiceConstants.LOG_TAG,
+                                    "MessageFrame",
+                                    "Received non-positive height value: ${height.toInt()}"
+                                )
```
prudrabhat, line 205: How do you feel about make this a little less verbose like this. That way we can identify what is wrong in the value received from the js layer while also falling back to the right value - 

```
// Declare this in the file.
private const val INVALID_HEIGHT = Integer.MIN_VALUE
```

```
onHeightReceived = { heightFromJs ->
     val adjustedHeight : Int =  heightFromJs?.toIntOrNull() ?: kotlin.run {
            Log.warning( ServiceConstants.LOG_TAG,  "MessageFrame", "Invalid height value received: $heightFromJs. Falling back to ${heightDp.value}")
            heightDp.value
       }

     heightDp.value = adjustedHeight.dp
}
                
```

code/core/src/test/java/com/adobe/marketing/mobile/services/ui/message/InAppMessageSettingsTest.kt
```diff
@@ -198,6 +198,7 @@ class InAppMessageSettingsTest {
             .backdropOpacity(0.5f)
             .cornerRadius(10.0f)
             .shouldTakeOverUi(true)
+            .setFitToContent(true)
```
prudrabhat, line 201: This is parameter is being prepared but not asserted anywhere in your test.

code/core/src/androidTest/java/com/adobe/marketing/mobile/services/ui/message/views/MessageScreenTestHelper.kt
```diff
@@ -28,7 +28,7 @@ internal object MessageScreenTestHelper {
      * Validates the size of the view bounds with the given height and width
      */
     internal fun validateViewSize(viewBounds: DpRect, height: Dp, width: Dp) {
-        viewBounds.height.assertIsEqualTo(height, "failed", Dp(2f))
+        viewBounds.height.assertIsEqualTo(height, "failed", Dp(16f))
```
prudrabhat, line 31: Any reason this test assertion was changed, despite no actual variation to the test case?

code/core/src/main/java/com/adobe/marketing/mobile/services/ui/message/InAppMessageSettings.kt
```diff
@@ -48,7 +48,8 @@ class InAppMessageSettings private constructor(
     val cornerRadius: Float,
     val shouldTakeOverUi: Boolean,
     val assetMap: Map<String, String>,
-    val gestureMap: Map<MessageGesture, String>
+    val gestureMap: Map<MessageGesture, String>,
+    val fitToContent: Boolean
```
prudrabhat, line 52: Since this is a new public api being added to the core services api surface  we will need to update the api dump file. This is [causing the PR check to fail](https://github.com/adobe/aepsdk-core-android/actions/runs/14494921620/job/40660375295?pr=748)
You can run `make api-dump` from your terminal to generate the updated api dump with this included.

prudrabhat (changes requested):

code/core/src/androidTest/java/com/adobe/marketing/mobile/services/ui/message/views/MessageScreenTestHelper.kt
spoorthipujariadobe: This is the tolerance of failure accepted. While a low value like 2 is acceptable due to rounding up and down when converting between pixel to dp, 16dp difference is not normal and might be an indication of existing logic breaking because of the new changes. @siddique-adobe please verify

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageFrame.kt
siddique-adobe: @prudrabhat this is not regressing the existing behaviour of the in app message height because heightDp state will change from here if the `fitToContent is disabled` but need to guard the same in case of `fitToContent is enabled` because that is actually overriding the heightDp state
for width i've not made any changes

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageFrame.kt
siddique-adobe: done @prudrabhat

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageFrame.kt
prudrabhat: Ah! Yes you are correct. I missed the negation in the beginning 😄. This makes sense.

code/core/src/phone/java/com/adobe/marketing/mobile/services/ui/message/views/MessageFrame.kt
siddique-adobe: No, This is not required for fit to content, because we're updating the `heightDp` value from `handleJavascriptMessage` listener, So in case of recomposition when the `fitToContent` is enabled we need to guard the recomposition to `with(density) { ((contentView.height.toDp() * inAppMessageSettings.height) / 100)`

code/core/src/androidTest/java/com/adobe/marketing/mobile/services/ui/message/views/MessageScreenTestHelper.kt
siddique-adobe: removed @spoorthipujariadobe and @prudrabhat

code/core/src/main/java/com/adobe/marketing/mobile/services/ui/message/InAppMessageSettings.kt
siddique-adobe: Done!, Thanks for help @prudrabhat

code/core/src/test/java/com/adobe/marketing/mobile/services/ui/message/InAppMessageSettingsTest.kt
siddique-adobe: Done @prudrabhat

spoorthipujariadobe:
Merged in my PR #753 to add support for max width to this PR. We can merge this branch to dev after we finish testing both features and ready for release

prudrabhat (approved):

PR #750, opened 2025-04-10
Fix ordered search type lookup in event history

spoorthipujariadobe:
## Description

The current ordered event history lookup logic uses newest timestamp of the previous event as the from value to search the existence of next event in the events array, instead of the oldest timestamp. Content card disqualification will be using ordered event look up and is failing because of the incorrect implementation.

## Related Issue

https://github.com/adobe/aepsdk-core-android/pull/749

## Motivation and Context

The `historical` rule condition allows checking if a series of events occurred in a specific order in the SDK event history by providing the `searchType` as `ordered` . As per the spec, for `ordered` search type, "_Each provided event will be queried in order.  As long as the count of an event is greater than one, the subsequent event will be queried for using the oldest timestamp of the previous event.  If the end of the events array is reached and each event has been found, the evaluation returns one (1)._"

This was implemented correctly in the [Java version](https://github.com/adobe/aepsdk-core-android/blame/0856ec47cf01a3650e48e7c170dd96ba3e80c2bd/code/android-core-library/src/main/java/com/adobe/marketing/mobile/internal/eventhub/history/AndroidEventHistory.java#L77-L124). However, when it was migrated to [Kotlin](https://github.com/adobe/aepsdk-core-android/blob/main/code/core/src/main/java/com/adobe/marketing/mobile/internal/eventhub/history/AndroidEventHistory.kt#L75-L123), we ended up using the [newest timestamp](https://github.com/adobe/aepsdk-core-android/blob/main/code/core/src/main/java/com/adobe/marketing/mobile/internal/eventhub/history/AndroidEventHistory.kt#L111) of the previous event as the from timestamp for the subsequent event. So the lookup would fail for the following case: [A, B, C, A] since we look for existence of B & C after the second occurrence of A, which is incorrect.

## How Has This Been Tested?

Added a test for ordered event history look up with multiple identical events

## Types of changes

## Checklist:

### DIFF
--- a/code/core/src/androidTest/java/com/adobe/marketing/mobile/internal/eventhub/history/AndroidEventHistoryTests.kt
+++ b/code/core/src/androidTest/java/com/adobe/marketing/mobile/internal/eventhub/history/AndroidEventHistoryTests.kt
@@ -88,6 +88,25 @@ class AndroidEventHistoryTests {
     }
 
     @Test
+    fun testGetEventsWithEnforceOrderWithMultipleIdenticalEvents() {
+        val data = mapOf("key" to "value")
+        assertTrue(record(data, 5000))
+
+        val data1 = mapOf("key1" to "value1")
+        assertTrue(record(data1, 10000))
+
+        assertTrue(record(data, 20000))
+
+        val requests = arrayOf(
+            EventHistoryRequest(data, 0, 0),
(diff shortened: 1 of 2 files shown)


yangyansong-adbe (approved):
LGTM

Issue #730, opened 2025-01-07
configureWithAppId does not refresh config in all scenarios


emdobrin:
### Expected Behaviour
If I need to test with two app ids in the same session, I should be able to switch between the two multiple times within that same session.

e.g.
MobileCore.configureWithAppId(appId1) -> set appId1
MobileCore.configureWithAppId(appId2) -> set appId2
MobileCore.configureWithAppId(appId1) -> set appId1

### Actual Behaviour

MobileCore.configureWithAppId(appId1) -> set appId1
MobileCore.configureWithAppId(appId2) -> set appId2
MobileCore.configureWithAppId(appId1) -> still **appId2** ❗

### Reproduce Scenario (including but not limited to)
Issue seems to be caused by [this config expiry check](https://github.com/adobe/aepsdk-core-android/blob/main/code/core/src/main/java/com/adobe/marketing/mobile/internal/configuration/ConfigurationExtension.kt#L247C40-L247C56) which caches the first appId1 set date and does not account for this scenario.

You might also need to consider clearing the expiry cache on configureWithAppId(null / "")

#### Platform and Version
Android, Core/Config v3.2.0

PR #755, opened 2025-05-05
Bug fix for the configureWithAppId API

yangyansong-adbe:
https://github.com/adobe/aepsdk-core-android/issues/730

### DIFF
--- a/code/core/src/main/java/com/adobe/marketing/mobile/internal/configuration/ConfigurationStateManager.kt
+++ b/code/core/src/main/java/com/adobe/marketing/mobile/internal/configuration/ConfigurationStateManager.kt
@@ -76,11 +76,8 @@ internal class ConfigurationStateManager {
     internal var environmentAwareConfiguration: Map<String, Any?> = mapOf()
         private set
 
-    /**
-     * Maintains a mapping between appId and the most recent download time when the config for that
-     * appId has been downloaded
-     */
-    private val configDownloadMap: MutableMap<String, Date> = mutableMapOf()
+    @VisibleForTesting
+    internal var lastDownloadedConfig: Pair<String, Date>? = null
 
     constructor(
         appIdManager: AppIdManager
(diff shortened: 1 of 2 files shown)


code/core/src/main/java/com/adobe/marketing/mobile/internal/configuration/ConfigurationStateManager.kt
```diff
@@ -76,11 +76,8 @@ internal class ConfigurationStateManager {
     internal var environmentAwareConfiguration: Map<String, Any?> = mapOf()
         private set
 
-    /**
-     * Maintains a mapping between appId and the most recent download time when the config for that
-     * appId has been downloaded
-     */
-    private val configDownloadMap: MutableMap<String, Date> = mutableMapOf()
+    @VisibleForTesting
+    internal var lastDownloadedConfig: Pair<String, Date>? = null
```
prudrabhat, line 80: I think the current version uses a tiny optimization to cache the older configs and check their validity. I am aligned that this may not be that important to keep in favor of the proposed method. 
However, can we make the same changes in iOS too to make it consistant.

yangyansong-adbe: Sure, if you agree this suggested approach won't break any current configuration workflows, I'll apply the same change on the iOS side as well.

emdobrin: @yangyansong-adbe and I had a quick chat on the topic, posting here a quick summary for posterity. 

We took a look at the internal logic for Configuration, and in addition to the `configDownloadMap` the previously downloaded configs are cached locally. Since we are not clearing these configs on switch, we found that this logic can give us an advantage for situations where network connectivity is not available/poor, and the app logic switches between two app ids at runtime. 
For example in the case where we set appIdA (1) -> appIdB (2) -> appIdA (3), appId (3) could still recover the cached config if the timeout did not expire and it was successfully cached on step 1.

yangyansong-adbe:
@emdobrin @prudrabhat  
I'm closing this PR and will create a new one with additional improvements we discussed, including using locally cached configurations in certain situations like poor network conditions.

PR #758, opened 2025-05-09
Catch runtime exceptions when retrieving the extended data from the intent

yangyansong-adbe:


code/core/src/main/java/com/adobe/marketing/mobile/internal/DataMarshaller.kt
```diff
@@ -62,26 +63,31 @@ internal object DataMarshaller {
...
-            }
-            try {
-                val value = extraBundle[key]
-                if (value?.toString()?.isNotEmpty() == true) {
-                    marshalledData[newKey] = value
+        try {
+            extraBundle.keySet()?.forEach { key ->
+                val newKey = when (key) {
+                    LEGACY_PUSH_MESSAGE_ID -> PUSH_MESSAGE_ID_KEY
+                    NOTIFICATION_IDENTIFIER_KEY -> LOCAL_NOTIFICATION_ID_KEY
```
emdobrin, line 73: We can extract our own keys before running the full loop, via direct key access or the Bundle#getString API. This way we can recover the keys we know, then try to read the remaining keys.

code/core/src/main/java/com/adobe/marketing/mobile/internal/DataMarshaller.kt
```diff
@@ -62,26 +63,31 @@ internal object DataMarshaller {
...
-            } catch (e: Exception) {
-                Log.error(CoreConstants.LOG_TAG, LOG_TAG, "Failed to retrieve data (key = $key) from Activity, error is: ${e.message}")
             }
-        }
 
-        extraBundle.remove(LEGACY_PUSH_MESSAGE_ID)
-        extraBundle.remove(NOTIFICATION_IDENTIFIER_KEY)
+            extraBundle.remove(LEGACY_PUSH_MESSAGE_ID)
+            extraBundle.remove(NOTIFICATION_IDENTIFIER_KEY)
+        } catch (e: Exception) {
```
emdobrin, line 88: For the try-catch blocks I recommend narrowing this down to the exception(s) you expect, in this case IllegalArgumentException.

code/core/src/androidTest/java/com/adobe/marketing/mobile/internal/DataMarshallerTests.kt
```diff
@@ -206,6 +209,19 @@ class DataMarshallerTests {
...
     }
+
+    @Test
+    fun marshalIntentExtras_whenBundleKeySetThrowsException_NoCrash() {
+
+        val mockIntent = Mockito.mock(Intent::class.java)
+        val mockBundle = Mockito.mock(Bundle::class.java)
+
+        `when`(mockIntent.extras).thenReturn(mockBundle)
+        `when`(mockBundle.keySet()).thenThrow(RuntimeException::class.java)
```
emdobrin, line 220: do we have any options to mimic a bad parcel write to get closer to the real behavior?

code/core/src/main/java/com/adobe/marketing/mobile/internal/DataMarshaller.kt
prudrabhat: @emdobrin , @yangyansong-adbe 
Given that the operation that is causing the exception is `extraBundle.keySet()`, may I recommend trying the best case first and then, in-case of an exception try to make a best effort retrieval of known keys as a fallback? That way we treat specific key extraction as a special case and easier code flow, testing.

code/core/src/androidTest/java/com/adobe/marketing/mobile/internal/DataMarshallerTests.kt
yangyansong-adbe: Nope

code/core/src/main/java/com/adobe/marketing/mobile/internal/DataMarshaller.kt
yangyansong-adbe: Either way works for me.

code/core/src/main/java/com/adobe/marketing/mobile/internal/DataMarshaller.kt
```diff
@@ -62,26 +63,57 @@ internal object DataMarshaller {
...
-                    marshalledData[newKey] = value
+
+        // First, handle known keys directly to avoid potential exceptions from keySet()
+        handleKnownKey(extraBundle, LEGACY_PUSH_MESSAGE_ID, PUSH_MESSAGE_ID_KEY, marshalledData)
+        handleKnownKey(extraBundle, NOTIFICATION_IDENTIFIER_KEY, LOCAL_NOTIFICATION_ID_KEY, marshalledData)
+
+        // Remove known keys from bundle after processing
+        try {
+            extraBundle.remove(LEGACY_PUSH_MESSAGE_ID)
+            extraBundle.remove(NOTIFICATION_IDENTIFIER_KEY)
```
emdobrin, line 77: Code readability improvements: 
- Would be nice to hold these keys in a known keys map
- Could we read and remove each key as part of handleKnownKey
- Please indicate in the api docs what parameters are mutated by handleKnownKey

prudrabhat: +1 adding the keys to an immutable map for lookup &  removing the known key from the bundle once handled. Looks like we always remove it irrespective of whether the handling is successful or not.

yangyansong-adbe: Yes, it's intended to remove that data as they're only used by our SDK.

code/core/src/main/java/com/adobe/marketing/mobile/internal/DataMarshaller.kt
```diff
@@ -24,10 +24,15 @@ internal object DataMarshaller {
...
-    private const val PUSH_MESSAGE_ID_KEY = "pushmessageid"
-    private const val NOTIFICATION_IDENTIFIER_KEY = "NOTIFICATION_IDENTIFIER"
-    private const val LOCAL_NOTIFICATION_ID_KEY = "notificationid"
+
+    /**
+     * Map of known keys that need to be transformed during marshalling.
+     * Key: original key in the bundle, Value: transformed key for marshalled data
+     */
+    private val KNOWN_KEYS_MAP = mapOf(
+        "adb_m_id" to "pushmessageid", // Legacy push message ID
```
prudrabhat, line 33: nit: Prefer keeping the constants and then adding them in the map.

prudrabhat (approved):

prudrabhat (approved):

PR #759, opened 2025-05-10
Change ExtensionApi getHistoricalEvents API to be on par with Core iOS

spoorthipujariadobe:
## Description

1. Added new `EventHistoryResult` class to represent the result of looking up an event in event history. It has count (number of times the event appears in history, oldest and new occurrence timestamps. 
2. Introduced a new `getHistoricalEvents` API in `ExtensionApi` and `ExtensionContainer` class. This new API queries the input array of `EventHistoryRequest` in event history in provided ordered or unordered way and returns an array of EventHistoryResult, one for each request in the input.
3. Deprecated existing `getHistoricalEvents` API in `ExtensionApi` and `ExtensionContainer` class and replaced all its internal usages with the newly added API.
4. Deprecated `EventHistoryResultHandler` which is used by existing `getHistoricalEvents` API since its signature is the same as `AdobeCallback`. Replaced all internal usages of it with `AdobeCallback`

## Related Issue

https://github.com/adobe/aepsdk-core-ios

## Motivation and Context

As a part of changes to event history for content card disqualification, we need to add a new search type [mostRecent](https://wiki.corp.adobe.com/display/ADMSMobile/Requalification+support) for the historical search. While implementing it I noticed a difference in the definition of the `getHistoricalEvents` API between the Android `ExtensionApi `and iOS `ExtensionRuntime` classes. In [iOS](https://github.com/adobe/aepsdk-core-ios/blob/main/AEPCore/Sources/eventhub/ExtensionRuntime.swift#L125), the API returns an array of [EventHistoryResult](https://github.com/adobe/aepsdk-core-ios/blob/main/AEPCore/Sources/eventhub/history/EventHistoryResult.swift) when we do non-ordered look up. The `EventHistoryResult` has a count for number of occurrences of each request event in event history along with the oldest and newest occurrence.

But in [Android](https://github.com/adobe/aepsdk-core-android/blob/main/code/core/src/main/java/com/adobe/marketing/mobile/ExtensionApi.java#L184-L188) it just returns a [EventHistoryResultHandler](https://github.com/adobe/aepsdk-core-android/blob/main/code/core/src/main/java/com/adobe/marketing/mobile/EventHistoryResultHandler.java) with a single integer value, indicating the total count of events found. For most recent search type, we do need the timestamp of found events too. Hence a new API is added in Android ExtensionApi to allow implementing most recent search and to have parity with iOS.

## How Has This Been Tested?

Updated and made required edits to all unit, functional and integration tests to replace the deprecated API with the new one. All tests continue to pass indicating no change in business logic.

## Types of changes

## Checklist:

- [x] My change requires a change to the documentation.


code/core/api/core.api
```diff
@@ -138,10 +155,11 @@ public abstract class com/adobe/marketing/mobile/ExtensionApi {
 	public abstract fun createSharedState (Ljava/util/Map;Lcom/adobe/marketing/mobile/Event;)V
 	public abstract fun createXDMSharedState (Ljava/util/Map;Lcom/adobe/marketing/mobile/Event;)V
 	public abstract fun dispatch (Lcom/adobe/marketing/mobile/Event;)V
+	public abstract fun getHistoricalEvents ([Lcom/adobe/marketing/mobile/EventHistoryRequest;ZLcom/adobe/marketing/mobile/AdobeCallback;)V
 	public abstract fun getHistoricalEvents ([Lcom/adobe/marketing/mobile/EventHistoryRequest;ZLcom/adobe/marketing/mobile/EventHistoryResultHandler;)V
 	public abstract fun getSharedState (Ljava/lang/String;Lcom/adobe/marketing/mobile/Event;ZLcom/adobe/marketing/mobile/SharedStateResolution;)Lcom/adobe/marketing/mobile/SharedStateResult;
 	public abstract fun getXDMSharedState (Ljava/lang/String;Lcom/adobe/marketing/mobile/Event;ZLcom/adobe/marketing/mobile/SharedStateResolution;)Lcom/adobe/marketing/mobile/SharedStateResult;
-	public abstract fun recordHistoricalEvent (Lcom/adobe/marketing/mobile/Event;Lcom/adobe/marketing/mobile/EventHistoryResultHandler;)V
+	public abstract fun recordHistoricalEvent (Lcom/adobe/marketing/mobile/Event;Lcom/adobe/marketing/mobile/AdobeCallback;)V
```
spoorthipujariadobe, line 162: Note: This API was added by me in a [previous PR](https://github.com/adobe/aepsdk-core-android/pull/749/files#diff-bf954d4a1545934d2e4eb2b4a3804c1b51f82da4db1dd9dd60bc20dd86a0d927R188-R202) on this feature branch that has not been released yet.

code/core/src/main/java/com/adobe/marketing/mobile/ExtensionApi.java
```diff
@@ -181,11 +181,26 @@ public abstract void createXDMSharedState(
...
+    @Deprecated
     public abstract void getHistoricalEvents(
             @NonNull EventHistoryRequest[] eventHistoryRequests,
             boolean enforceOrder,
             @NonNull EventHistoryResultHandler<Integer> handler);
 
+    /**
+     * Retrieves a count of historical events matching the provided requests.
+     *
+     * @param eventHistoryRequests an array of {@link EventHistoryRequest}s used to generate the *
```
yangyansong-adbe, line 193: `used to generate the *` -> `used to generate the`

code/core/src/main/java/com/adobe/marketing/mobile/ExtensionApi.java
```diff
@@ -181,11 +181,26 @@ public abstract void createXDMSharedState(
...
+     *     hash and timeframe for the event lookup
+     * @param enforceOrder if `true`, consecutive lookups will use the oldest timestamp from the *
```
yangyansong-adbe, line 195: `use the oldest timestamp from the *` -> `use the oldest timestamp from the`

code/core/src/main/java/com/adobe/marketing/mobile/ExtensionApi.java
```diff
@@ -181,11 +181,26 @@ public abstract void createXDMSharedState(
...
+     *     previous event as their from date
+     * @param handler contains an {@link EventHistoryResult} for each provided request
```
yangyansong-adbe, line 197: an array of {@link EventHistoryResult}

code/core/src/main/java/com/adobe/marketing/mobile/internal/eventhub/ExtensionContainer.kt
```diff
@@ -278,15 +281,27 @@ internal class ExtensionContainer constructor(
...
     }
 
-    override fun recordHistoricalEvent(event: Event, handler: EventHistoryResultHandler<Boolean>) {
+    override fun getHistoricalEvents(
+        eventHistoryRequests: Array<out EventHistoryRequest>,
+        enforceOrder: Boolean,
+        handler: AdobeCallback<Array<EventHistoryResult>>
+    ) {
+        // TODO: What should be returned here if eventHistory is null - null or empty array?
+        EventHub.shared.eventHistory?.getEvents(eventHistoryRequests, enforceOrder, handler) ?: handler.call(null)
```
yangyansong-adbe, line 301: I vote for `empty array`

code/core/src/main/java/com/adobe/marketing/mobile/ExtensionApi.java
```diff
@@ -195,9 +210,9 @@ public abstract void getHistoricalEvents(
...
      *
      * @param event the {@link Event} to be recorded in the Event History database
-     * @param handler the {@link EventHistoryResultHandler} which returns a boolean indicating a
-     *     successful database insert
+     * @param handler the {@link AdobeCallback} which returns a boolean indicating a successful
+     *     database insert
      */
     public abstract void recordHistoricalEvent(
-            @NonNull final Event event, @NonNull final EventHistoryResultHandler<Boolean> handler);
+            @NonNull final Event event, @NonNull final AdobeCallback<Boolean> handler);
```
yangyansong-adbe, line 217: Q: what actions should we take if the underlying database operation encounters an error?  Logging errors? Retry? 
If we want to understand the error types before taking action, we can consider using `AdobeCallbackWithError`
Same question for the getter API.

code/core/src/main/java/com/adobe/marketing/mobile/launch/rulesengine/HistoricalEventsQuerying.kt
```diff
@@ -48,3 +59,43 @@ internal fun historicalEventsQuerying(
...
+ */
+@JvmSynthetic
+internal fun convertEventHistoryResultToInt(
+    enforceOrder: Boolean,
+    eventHistoryResult: Array<EventHistoryResult>
+): Int {
+    if (enforceOrder) {
+        for (result in eventHistoryResult) {
+            // Early exit on database error
+            if (result.count == -1) {
```
prudrabhat, line 80: 0's and 1's here are a bit too involved for code readability. Can we add designated constants for database error, all events found , no records found etc and use them across the calling components?

code/core/src/main/java/com/adobe/marketing/mobile/ExtensionApi.java
spoorthipujariadobe: I discussed this offline with Yansong and changed the callback to accept `AdobeCallbackWithError`. However for the time being I will still return `-1` in case of database error to ensure we're on par with iOS and avoid issue where the value is directly used to compare with a rule `matcher`.
We can choose to propagate errors together in both platforms in the future.

code/core/src/main/java/com/adobe/marketing/mobile/EventHistoryResult.kt
```diff
@@ -0,0 +1,24 @@
...
+package com.adobe.marketing.mobile
+
+/**
+ * Passed to handler of [ExtensionApi.getHistoricalEvents] API to represent the result of event look up in EventHistoryDatabase
+ * @param count The number of occurrences of the event in the event history database
+ * @param oldestOccurrence The time of the oldest occurrence of the event in the event history database. If count is 0, this value will be null
+ * @param newestOccurrence The time of the newest occurrence of the event in the event history database. If count is 0, this value will be null
+ */
+data class EventHistoryResult(
+    val count: Int,
```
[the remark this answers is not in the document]
spoorthipujariadobe: We can have it as an optional and set it to null. What do you think?

code/core/src/main/java/com/adobe/marketing/mobile/internal/eventhub/history/AndroidEventHistory.kt
```diff
@@ -131,29 +114,29 @@ internal class AndroidEventHistory : EventHistory {
...
      *
      * @param eventHistoryRequests an array of `EventHistoryRequest`s to be deleted
-     * @param handler a callback which will be called with a `int` containing the total number
+     * @param callback which will be called with a `int` containing the total number
      * of rows deleted from the `AndroidEventHistoryDatabase`
      */
     override fun deleteEvents(
         eventHistoryRequests: Array<out EventHistoryRequest>,
-        handler: EventHistoryResultHandler<Int>?
+        callback: AdobeCallbackWithError<Int>?
```
yangyansong-adbe, line 122: When database errors occur, the `AndroidEventHistoryDatabase.delete` function suppresses the exception and returns a `0`. Should we standardize it with the other get and record functions that return a `-1` to signify a database error?

spoorthipujariadobe: Good catch I will change it to return -1

spoorthipujariadobe: I checked with iOS implementation of this. Looks like even there `0` is returned in case of database error which means no records were deleted from event history. We might want to either keep this as is or change it on both platforms. Thoughts?
https://github.com/adobe/aepsdk-core-ios/blob/main/AEPCore/Sources/eventhub/history/EventHistoryDatabase.swift#L148-L151

yangyansong-adbe: I see. Given that we don't have a plan to retry today when the database errors occur, I think it should be fine to leave it as it is. Can you update the function docs to reflect this expected behavior?
@prudrabhat  Let us know if you have other thoughts.

code/core/src/main/java/com/adobe/marketing/mobile/ExtensionApi.java
prudrabhat: @spoorthipujariadobe , @yangyansong-adbe : After looking at the changes resulting from this comment, I feel that the partial use of `AdobeCallbackWithError`, that is, using the value in `call()` to notify both success error instead of `fail()`, will cause more confusion and errors than it is beneficial.

Further, I would not treat this as a seamless backward compatible contract as:
1. Current consumers who don't act/implement the `fail()` call will miss the notification because they expect the callback from `call()`. In order not to break them we would need to notify both the handler methods.
2. New consumers who act/implement the `fail()` call will see the notification twice because of point1 and they need to gate the clauses in their logic not to act on some codes in `call()` so as to exclusively handle them in `fail()` _or_ totally ignore the `fail()`(which would make introduction of`AdobeCallbackWithError` an unused change)

For these reasons, I recommend you to consider either  
a. use the `AdobeCallbackWithError` to truly propagate errors via `fail()`.  
b. use `AdobeCallback` and embed the error via `call()` to be similar to iOS. 

Reg (a) I looked at the `AdobeError` class here and unfortunately it is too generic for us to propagate meaningful database errors without adding a lot of new error codes. So I am leaning towards (b) due to the current limitations.

yangyansong-adbe: - `AdobeError` was created to be expandable, allowing for the inclusion of additional error codes. Initially, I suggested adding new error codes to reflect database errors.
- About `AdobeCallbackWithError`:  Since we're introducing new internal APIs, adopting the `AdobeCallbackWithError` interface now would be beneficial to avoid additional breaking changes later on if we ultimately use the `fail` function for error handling.

I agree that if we don't refactor both platforms for error handling, like retry, using `AdobeCallbackWithError` might be confusing. 
I personally prefer option A; it will make things simple and clear.

spoorthipujariadobe: Accepting `AdobeCallbackWithError` was mostly about future proofing. We still need to return `-1` instead of sending back the errors today to ensure platform consistency.

I was looking at a few other extensions that accept `AdobeCallback`. Since `AdobeCallbackWithError` extends `AdobeCallback`, they check to see the type of the callback passed and call the `fail` method accordingly. Here are examples from Target and Optimize SDKs
https://github.com/adobe/aepsdk-target-android/blob/main/code/target/src/phone/java/com/adobe/marketing/mobile/Target.java#L454-L457
https://github.com/adobe/aepsdk-optimize-android/blob/v3.0.0/code/optimize/src/main/java/com/adobe/marketing/mobile/optimize/Optimize.java#L319-L330

Wondering if this is an option to have this event history API accept `AdobeCallback` but still allow customers to listen for errors if we choose to propagate it in the future.

yangyansong-adbe: Initially, we use `AdobeCallback ` in public APIs. Later, we had to internally return an `AdobeCallbackWithError` object, mainly to address the customer's request to handle specific errors. It's only a workaround because we could not modify the public APIs. 
If we don't want to change a lot of code to return specific errors today, we can use `AdobeCallback`.

prudrabhat (approved):
LGTM. Please include changes suggested in my prev comment.

code/core/src/main/java/com/adobe/marketing/mobile/internal/eventhub/history/AndroidEventHistory.kt
spoorthipujariadobe: I changed this to call `fail` if an error occurs

yangyansong-adbe (approved):
LGTM

PR #767, opened 2025-06-07
Rules : Fix logical operation for negative matchers with multiple values #1147

PravinPK:
### **Problem**
 When processing multiple values, the rules engine used OR logic for all matchers , which creates a logical flaw for negative 
conditions. For example 
```
carriername != "AT&T" OR  carriername != "Verizon" OR carriername != "Cricket"
```

```
                                    {
                                      "definition": {
                                          "type": "matcher",
                                          "matcher": "ne",
                                          "key": "xdm.carriername"
                                          "values": [
                                              "AT&T",
                                              "Verizon",
                                             "Cricket"
                                           ],
                                       },
                                    },
```

This is always true for negative matchers! If carriername = "AT&T", the condition carriername != "Verizon" makes the entire OR expression true, causing incorrect rule firing.

###  **Impact**
This was causing rules to fire when they shouldn't. (Customer issue : IAM was fired too often)

### **Solution**
Modified JSONRulesParser.swift to use AND logic for negative matchers like notContains (nc) and notExists (ne).

Correct evaluation
```
carriername != "AT&T" AND carriername != "Verizon" AND carriername != "Cricket"
```

### DIFF
--- a/code/core/src/main/java/com/adobe/marketing/mobile/launch/rulesengine/json/MatcherCondition.kt
+++ b/code/core/src/main/java/com/adobe/marketing/mobile/launch/rulesengine/json/MatcherCondition.kt
@@ -43,7 +43,7 @@ internal class MatcherCondition(private val definition: JSONDefinition) : JSONCo
             "ex" to "exists",
             "nx" to "notExist"
         )
-        
+
         // Negative matchers that should use AND logic for multiple values
         private val NEGATIVE_MATCHERS = setOf("ne", "nc")
     }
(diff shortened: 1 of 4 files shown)


code/core/src/main/java/com/adobe/marketing/mobile/launch/rulesengine/json/MatcherCondition.kt
```diff
@@ -42,6 +43,9 @@ internal class MatcherCondition(private val definition: JSONDefinition) : JSONCo
             "ex" to "exists",
             "nx" to "notExist"
         )
+
+        // Negative matchers that should use AND logic for multiple values
+        private val NEGATIVE_MATCHERS = setOf("ne", "nc")
```
prudrabhat, line 48: Should we also include `notExist` here?

PravinPK:
Thanks @prudrabhat for getting the branches right

PR #790, opened 2025-08-15
Fix race condition in rules engine which caused some events to be processed twice

praveek:
Issue: https://github.com/adobe/aepsdk-core-android/issues/678
Fixes an issue where cached events occurring between replaceRules and the resetRules event being received by the rules engine are processed twice.


yangyansong-adbe (approved):

prudrabhat (approved):

praveek:
Closing this for now, will reopen after extension tests.

PR #825, opened 2026-08-31
refactor: refactored log statements

### LINKED ISSUE #806: SQiteDataQueue inefficiently calling DataEntity.toString when trace is disabled
[https://github.com/adobe/aepsdk-core-android/blob/6353dd565267c571297a2b636973738377b921a6/code/cor…](https://github.com/adobe/aepsdk-core-android/blob/6353dd565267c571297a2b636973738377b921a6/code/core/src/phone/java/com/adobe/marketing/mobile/services/SQLiteDataQueue.java#L184)
 
We noticed a low memory report that referenced this line.  It is converting the entity to a `String` even when trace logging is not enabled.
 
```
            Log.trace("Services", this.LOG_PREFIX, String.format("peek - Successfully returned DataEntity (%s)", ((DataEntity)dataEntities.get(0)).toString()), new Object[0]);
            return (DataEntity)dataEntities.get(0);
 
```
 
In addition the log line seems like it should be able to avoid the manual `toString()` call since `String.format` is being used.
 
Could it use this function instead to delay the string formatting? `Log.trace(@NonNull String extension, @NonNull String source, @NonNull String format, Object... params)`

shushinde:
While looking into https://github.com/adobe/aepsdk-core-android/issues/806, I noticed the same pattern all over the core services: log messages get built with String.format right at the call site, so the string (and any toString()) is created even when that log level is turned off.

This PR moves all of those calls to Log's varargs overload, so the format string and its arguments are passed straight through and only formatted if the log level is actually enabled. That covers the peek() case from the issue plus 50 other spots across 8 files.

It's a mechanical change, no behavior or log output change; just less wasted work on the hot path.

I ran spotlessCheck, compiled :core, and all 1143 core unit tests pass. Closes https://github.com/adobe/aepsdk-core-android/issues/806

### DIFF
--- a/code/core/src/main/java/com/adobe/marketing/mobile/internal/util/SQLiteDatabaseHelper.java
+++ b/code/core/src/main/java/com/adobe/marketing/mobile/internal/util/SQLiteDatabaseHelper.java
@@ -43,10 +43,10 @@ public class SQLiteDatabaseHelper {
             Log.warning(
                     CoreConstants.LOG_TAG,
                     LOG_PREFIX,
-                    String.format(
-                            "createTableIfNotExists - Error in creating/accessing database (%s)."
-                                    + "Error: (%s)",
-                            dbPath, e.getMessage()));
+                    "createTableIfNotExists - Error in creating/accessing database (%s)."
+                            + "Error: (%s)",
+                    dbPath,
+                    e.getMessage());
             return false;
         } finally {
             closeDatabase(database);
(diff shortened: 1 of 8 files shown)


sagar-sharma-adobe (approved):

shushinde:
mistakenly raise for main, 
from log refactor -> staging https://github.com/adobe/aepsdk-core-android/pull/826

sagar-sharma-adobe (approved):

PR #830, opened 2026-09-15
Fix DataMarshaller OOM: avoid unnecessary toString() calls on Intent extras

### LINKED ISSUE #791: DataMarshaller repeatedly calls toString() on extras causing memory pressure
`DataMarshaller` repeatedly calls `toString()` on objects that are not used. If an `Activity.intent` holds `Parcelable` Extras it is automatically captured by `MobileCore.collectLaunchInfo`.  We have noticed the following lines in OutOfMemory stack traces within crash reports where it is unable to allocate memory for large strings.  Is it possible to avoid calling `toString()` on `Parcelable` extras in an `Intent`, especially since the return value is only used to check if the object is empty?
 
> OBFUSCATED.toString (OBFUSCATED:140)
> java.lang.String.valueOf (String.java:3657)
> java.lang.StringBuilder.append (StringBuilder.java:132)
> java.util.AbstractCollection.toString (AbstractCollection.java:473)
> java.lang.String.valueOf (String.java:3657)
> java.lang.StringBuilder.append (StringBuilder.java:132)
> OBFUSCATED.toString (OBFUSCATED:128)
> com.adobe.marketing.mobile.internal.DataMarshaller.marshalIntentExtras$core_phoneRelease (DataMarshaller.kt:93)
> com.adobe.marketing.mobile.internal.DataMarshaller.marshal (DataMarshaller.kt:63)
> com.adobe.marketing.mobile.MobileCore.collectLaunchInfo (MobileCore.java)
> com.adobe.marketing.mobile.LaunchInfoCollector.onActivityResumed (MobileCoreInitializer.kt:243)
> com.adobe.marketing.mobile.services.internal.context.App$ActivityTracker.onActivityResumed (App.kt:173)
> com.adobe.marketing.mobile.services.internal.context.App$InternalActivityLifecycleCallbacks.onActivityResumed (App.kt:186)
> android.app.Application.dispatchActivityResumed (Application.java:431)
> android.app.Activity.dispatchActivityResumed (Activity.java:1445)
 
    internal fun marshalIntentExtras(intent: Intent, marshalledData: MutableMap<String, Any>) {
        val extraBundle = intent.extras ?: return
 
        // First, handle known keys directly to avoid potential exceptions from keySet()
        KNOWN_KEYS_MAP.forEach { (originalKey, transformedKey) ->
            processAndRemoveKnownKey(extraBundle, originalKey, transformedKey, marshalledData)
        }
 
        // Then extract other keys
        try {
            extraBundle.keySet()?.forEach { key ->
                try {
                    if (KNOWN_KEYS_MAP.containsKey(key)) {
                        // returning from a lambda expression, the loop will continue.
                        return@forEach
                    }
                    val value = extraBundle[key]
                    if (value?.toString()?.isNotEmpty() == true) { // toString() can create large String objects that are unused in the rest of this method
                        marshalledData[key] = value
                    }
                } catch (e: Exception) {
                    Log.error(CoreConstants.LOG_TAG, LOG_TAG, "Failed to retrieve data (key = $key) from Activity, error is: ${e.message}")
                }
            }
        } catch (e: Exception) {
            Log.error(CoreConstants.LOG_TAG, LOG_TAG, "Failed to retrieve data from Activity, error is: ${e.message}")
        }
    }

harsheenkohli:
## Summary
- `DataMarshaller.marshalIntentExtras()` called `value.toString()` on every Intent extra purely to check non-emptiness. For large `Parcelable`s, collections, or nested `Bundle`s, this forced full stringification of the object graph just to discard the result — causing `OutOfMemoryError` crashes on every `Activity` resume (see crash trace in #791).
- Replaces that check with `isMarshallable(value)`, combining a type-aware compatibility check (`isEventDataCompatible`) and a non-allocating emptiness check (`isEffectivelyEmpty`) in a single function, so a caller only needs one call instead of remembering to check both conditions.
- Exposes `EventDataUtils.isImmutableType()` so `DataMarshaller`'s compatibility check delegates to the actual canonical type set instead of maintaining its own duplicate whitelist — a new Event-data-compatible type now only needs to be added in one place.
- Fixes stale javadoc on `Event.Builder.setEventData()` that omitted `BigDecimal`, `BigInteger`, `Character`, `UUID`, and arrays from the list of accepted value types.

Root-cause trace and design rationale documented here: [wiki page](https://wiki.corp.adobe.com/pages/viewpage.action?pageId=4041036674).

Fixes #791

## Test plan
- [x] Added `marshalIntentExtras_largeCollectionExtra_doesNotCallToString` — 200k-entry collection wrapped in a `toString()`-call-counting spy; confirms `toString()` is never invoked and the value still survives marshalling.
- [x] Added `marshalIntentExtras_unsupportedTypeExtra_isFilteredOut` — a plain `Serializable` type outside the supported set; confirms it's filtered out in `DataMarshaller` rather than only being dropped later downstream by `EventDataUtils.immutableClone()`.
- [x] Full `DataMarshallerTests` suite passes (17/17) on-device.
- [x] Full unit test suite passes (1143/1143), including `EventDataUtilsTests`.


code/core/src/main/java/com/adobe/marketing/mobile/internal/DataMarshaller.kt
```diff
@@ -102,6 +103,53 @@ internal object DataMarshaller {
...
     }
 
+    /**
+     * True if [value] should be carried forward into marshalled Intent data: it must be
+     * representable in Event data (see [isEventDataCompatible]) and not effectively empty
+     * (see [isEffectivelyEmpty]). Combining both checks in one place means a type added to
+     * one side of the check doesn't require a second matching change elsewhere.
+     */
+    private fun isMarshallable(value: Any): Boolean =
+        isEventDataCompatible(value) && !isEffectivelyEmpty(value)
```
sagar-sharma-adobe: Let's move the implementation of both isEventDataCompatible and isEffectivelyEmpty to EventDataUtils to make sure that the same logic and types are being used and reduce redundancy.

sagar-sharma-adobe:
Looks good, just one refactor requested.

code/core/src/main/java/com/adobe/marketing/mobile/internal/DataMarshaller.kt
harsheenkohli: Done, moved both isEventDataCompatible and isEffectivelyEmpty into EventDataUtils:                 
<https://github.com/adobe/aepsdk-core-android/commit/0f1a1bc998928fa74cddc97706e6ac28a4598418>

sagar-sharma-adobe (approved):

